PR #17, opened 2023-05-15
Merkle Sum Tree optimization

enricobottazzi:
It optimizes the generation of a merkle sum tree thanks to parallelization using `thread` standard create.

A write up for the process of improving the performance of the MST can be found [here](https://hackmd.io/fLJHxNToQRq03ZOxvxJ80A) and applied at 0c34b73. 

Further optimization involves:
- the `tree` object that has been made more memory efficient at 02ed0ab
- the `compute_leaves` function uses a more efficient poseidon hasher that takes 2 input at bae6fc2. The previous version would pad the input with zeroes


.gitignore
```diff
@@ -1,3 +1,9 @@
+# Generated by Cargo
+# will have compiled files and executables
 /target
 /ptau
-/benches/csv
\ No newline at end of file
+/benches/csv
+
+# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
```
alxkzmn: I'd say it doesn't hurt to have it checked in into Git to prevent any potential issues. Do you imagine the crate being used as a dependency by someone in the future?

alxkzmn (approved):

.gitignore
enricobottazzi: I still don't know how it will be consumed. But I agree with you that it is better to stay on the safe side for now. Removing it from `.gitignore`

sifnoc (approved):

PR #22, opened 2023-05-30
Multiasset MST

### LINKED ISSUE #18: Add multi-asset support
The Merkle sum tree should support multiple assets

alxkzmn:
Resolves #18 by adding multi-asset support to the MST. Added the python scripts that generate the necessary Poseidon hasher parameters according to the set number of assets.


enricobottazzi:
One minor logistic thing: 

I removed this PR from the Chip readiness milestone as we already had the issue with the milestone tag and you already, correctly, had attached this PR to the related issue

PR #33, opened 2023-06-02
Integrated Overflow chip on MST 

sifnoc:
There are two practical configurations, which are tested, for the overflow chip:

1. 12 bits with 21 Columns
2. 8 bits with 31 Columns

All test cases within the Merkle Sum Tree have been modified to accommodate the second overflow chip configuration.


src/chips/merkle_sum_tree.rs
```diff
@@ -287,7 +293,31 @@ impl MerkleSumTreeChip {
...
+            ],
+        )?;
+
+        // Initiate the overflow check chip
+        let overflow_chip = OverflowChip::construct(self.config.overflow_check_config.clone());
+
+        overflow_chip.load(&mut layouter)?;
+
+        // Each balance cell is constrained to be less than the maximum limit
+        overflow_chip.assign(
```
enricobottazzi, line 310: I think that there's an issue here. There's no constraint in place that forces `update_value` to be equal to `left_balance` or to `right_balance`. I think this can be resolved by modifing assign function inside the overflow_chip to take `update_value: &AssignedCell<Fp, Fp>` as input and copying the assigned cell to `self.config.a`. This is a similar operation to the one performed in the `enforce_less_than()` function.

src/chips/overflow/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod overflow_check;
+pub mod tests;
```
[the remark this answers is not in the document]
sifnoc: Fixed at [d00ed53](https://github.com/summa-dev/summa-solvency/pull/33/commits/d00ed534c4f959a33662ea6a41664e2b764599d5)

src/chips/merkle_sum_tree.rs
```diff
@@ -287,7 +295,31 @@ impl MerkleSumTreeChip {
...
+            layouter.namespace(|| "overflow check left balance"),
+            left_balance.value().map(|x| x.to_owned()),
+        )?;
+        overflow_chip.assign(
+            layouter.namespace(|| "overflow check right balance"),
+            right_balance.value().map(|x| x.to_owned()),
+        )?;
+        overflow_chip.assign(
+            layouter.namespace(|| "overflow check computed sum"),
```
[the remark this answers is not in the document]
sifnoc: Yes, good point! we do not need to check overflow at `computed_sum`.
Updated at [d00ed53](https://github.com/summa-dev/summa-solvency/pull/33/commits/d00ed534c4f959a33662ea6a41664e2b764599d5)

src/chips/overflow/tests.rs
```diff
@@ -0,0 +1,259 @@
...
+            0x0fffffffffffffff,
+        ]));
+
+        let circuit = OverflowCheckCircuit::<8, 31> {
+            a: max_balance,
+            b: max_balance,
+        };
+        let invalid_prover = MockProver::run(k, &circuit, vec![]).unwrap();
+
+        fn gen_errors(region_num: usize, fisrt_advice: &str, last_advice: &str) -> VerifyFailure {
```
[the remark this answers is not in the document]
sifnoc: Fixed at [d00ed53](https://github.com/summa-dev/summa-solvency/pull/33/commits/d00ed534c4f959a33662ea6a41664e2b764599d5)

src/chips/merkle_sum_tree.rs
```diff
@@ -94,9 +101,7 @@ impl MerkleSumTreeChip {
             vec![s * (left_balance + right_balance - computed_sum)]
         });
 
-        let hash_inputs = (0..WIDTH).map(|_| meta.advice_column()).collect::<Vec<_>>();
-
-        let poseidon_config = PoseidonChip::<Spec4, WIDTH, RATE, L>::configure(meta, hash_inputs);
+        let poseidon_config = PoseidonChip::<Spec4, WIDTH, RATE, L>::configure(meta);
```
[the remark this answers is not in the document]
sifnoc: Fixed at [b19d029](https://github.com/summa-dev/summa-solvency/pull/33/commits/b19d029acac584127dfb34474b1ba918ad2e174c) also update at [d00ed53](https://github.com/summa-dev/summa-solvency/pull/33/commits/d00ed534c4f959a33662ea6a41664e2b764599d5)

src/merkle_sum_tree/utils/csv_parser.rs
```diff
@@ -26,12 +26,12 @@ pub fn parse_csv_to_entries<P: AsRef<Path>>(path: P) -> Result<Vec<Entry>, Box<d
         entries.push(entry);
     }
 
-    // modulus from bn256 curve impl => https://github.com/privacy-scaling-explorations/halo2curves/blob/main/src/bn256/fr.rs#L38
-    const MODULUS_STR: &str = "30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001";
+    // For preventing overflow, we set the maximum value limit to 2^251 - 1
```
enricobottazzi, line 29: Why not 252? I think we can get there using 84*3.

src/chips/merkle_sum_tree.rs
```diff
@@ -1,9 +1,12 @@
+use crate::chips::overflow::overflow_check::{OverflowCheckConfig, OverflowChip};
 use crate::chips::poseidon::hash::{PoseidonChip, PoseidonConfig};
 use crate::chips::poseidon::spec::Spec4;
 use gadgets::less_than::{LtChip, LtConfig, LtInstruction};
 use halo2_proofs::halo2curves::bn256::Fr as Fp;
 use halo2_proofs::{circuit::*, plonk::*, poly::Rotation};
 
+const ACC_COLS: usize = 31;
```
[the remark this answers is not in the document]
sifnoc: I agree that minimizing the number of columns is beneficial, but increasing the number of bits to reduce columns in turn requires more rows for the fixed column. This implies a longer time for generating proofs. For instance, setting 20 Bits and 10 Columns resulted in a 9141s test time for `test_valid_merkle_sum_tree`. 
Therefore, practically speaking, setting the `MAXBITS` to less than 18 appears to be the maximum choice for now.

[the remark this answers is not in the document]
enricobottazzi: Oh damn, I see it now! 

Ok let's keep it like this for now then

src/merkle_sum_tree/utils/csv_parser.rs
sifnoc: Currently, we're using a single range table where `1 << 8` is applied to the decomposed columns for range checking. This means the overflow chip checks if the maximum number is (1 << 251) - 1. The following table might be useful for understanding:

|   	| a_0                                                            	| a_1  	| a_2  	| ... 	| a_30 	|
|---	|----------------------------------------------------------------	|------	|------	|-----	|------	|
| x 	| 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff 	| 0xff 	| 0xff 	| ... 	| 0xff 	|

In an ideal scenario, we could set the maximum value to half the modulus to avoid overflow, irrespective of the decomposed value. However, to achieve this, we'd need to create tables for range checking for each decomposed column.

|   	| a_0                                                                	| a_1  	| a_2  	| ... 	| a_32 	|
|---	|--------------------------------------------------------------------	|------	|------	|-----	|------	|
| x 	| 0x183227397098d014dc2822db40c0ac2e9419f4243cdcb848a1f0fac9f8000000 	| 0x18 	| 0x32 	| ... 	| 0x00 	|

However, I have my doubts about this approach. Each column from `a_1` to `a_30` requires a different range table. a_1 uses 0..24(0x18), a_2 uses 0..50(0x32), and so on.

One alternative that seems plausible is to use `0x1832` as the range for the table. While we could go for a larger number, we would need to set the maximum value to 2^n to fill all 252 bits. Additionally, as mentioned earlier, it seems challenging to exceed `0x183227`, which is between `1 << 20` and `1 << 21`.

The following table demonstrates how the maximum value can be checked for overflow using the 0x1832 range table:

|   	| a_0                                                                	| a_1    	| a_2    	| ... 	| a_17   	|
|---	|--------------------------------------------------------------------	|--------	|--------	|-----	|--------	|
| x 	| 0x1832183218321832183218321832183218321832183218321832183218321832 	| 0x1832 	| 0x1832 	| ... 	| 0x1832 	|

What are your thoughts on turning this idea into a new task?

src/chips/merkle_sum_tree.rs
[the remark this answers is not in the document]
enricobottazzi: @sifnoc I found a good spot setting 

```
const ACC_COLS: usize = 20;
const MAX_BITS: u8 = 10;
```

Set k to 13 and run 

```
cargo  test --release test_valid_merkle_sum_tree_with_full_prover  -- --nocapture
Start:   Creating proof
End:     Creating proof ............................................................2.137s
```

I still think that it is ok to keep it as it for now and later try tweaking for a good balance between proving time and verification cost. Just sharing it for your interest

src/chips/overflow/overflow_check.rs
```diff
@@ -39,14 +39,36 @@ impl<const MAX_BITS: u8, const ACC_COLS: usize> OverflowChip<MAX_BITS, ACC_COLS>
...
+                // A single range table is utilized which applies `1 << 8` to decompose the columns for range checking.
+                //
+                // Consider the example where ACC_COLS = 3, the decomposed values would be represented as follows:
+                //
+                // |     | a_0 (value) | a_1  | a_2  | a_3  |
+                // |-----|-------------|------|------|------|
+                // |  x  | 0xffffff    | 0xff | 0xff | 0xff |
+                //
+                // Here, each column `a_n` represents a decomposed value.
+                // So, decomposed_value_sum would be calculated as a_0 * 2^16 + a_1 * 2^8 + a_2 * 1.
```
enricobottazzi, line 62: Thanks for the refactoring. I think there's a typo here in the comment as I guess you meant 

```
a_1 * 2^16 + a_2 * 2^8 + a_3* 1.
```

Also, just to be sure, the decomposition in performed little endian, namely the most significant bits is in the first advice column?

src/merkle_sum_tree/utils/csv_parser.rs
enricobottazzi: The idea of setting this check 

```
    // throw error if balance is larger than the modulus
    if balance_acc >= BigInt::parse_bytes(MODULUS_STR.as_bytes(), 16).unwrap() {
        return Err("Balance is larger than the modulus".into());
    }
```

was to warn the CEX that if they are building a MST which balance sum is larger than the modulus, it will end up in an overflow during proof generation.

Now I see that it would be more appropriate to check that the balance should not be larger than `MAX_VALUE_STR` which is the actual range check that we perform inside the circuit for the balance. 

But the value `MAX_VALUE_STR` depends on the value assigned to `ACC_COLS` and `MAX_BITS` which are generic to the circuit and might vary. Therefore I think it makes sense to keep this check against `MODULUS_STR`.  

In the case that `MODULUS < balance_acc < MAX_VALUE` the cex would be able to build the merkle sum tree and incur in an error during proof generation, but I think this is fine.


> In an ideal scenario, we could set the maximum value to half the modulus to avoid overflow, irrespective of the decomposed value

I didn't understand this part

src/chips/merkle_sum_tree.rs
sifnoc: Ah, indeed! I completely overlooked this section! Thanks for bringing it to my attention
 
Fixed at [4a42627](https://github.com/summa-dev/summa-solvency/pull/33/commits/4a42627ac6bb8b1d628dc5ca9ac2deedc4d862ac)

src/merkle_sum_tree/utils/csv_parser.rs
sifnoc: > But the value `MAX_VALUE_STR` depends on the value assigned to `ACC_COLS` and `MAX_BITS` which are generic to the circuit and might vary. Therefore I think it makes sense to keep this check against `MODULUS_STR`.
> 
> In the case that `MODULUS < balance_acc < MAX_VALUE` the cex would be able to build the merkle sum tree and incur in an error during proof generation, but I think this is fine.

I agree with that, it doesn't seem like the right time to settle this. Let's revert to this.

> > In an ideal scenario, we could set the maximum value to half the modulus to avoid overflow, irrespective of the decomposed value
> 
> I didn't understand this part

If we allow balances to go beyond half the modulus, meaning the overflow chip range check exceeds half the modulus, we could run into a worst-case scenario like the one below (considering a modulus of 255 for simplicity):

Let's walk through the 'merkle_prove_layer' in the circuit with some numbers:
https://github.com/summa-dev/summa-solvency/blob/b0fa4163474ac8ee1988342e070221f5586fb7a9/src/circuits/merkle_sum_tree.rs#L58-L81

leaf_balance = 28
path_element_balance[0] = 100, next_sum = 128
path_element_balance[1] = 128, next_sum = 1

In such a situation, the final next_sum value becomes 1, which wouldn't prevent an overflow situation.

This is the rationale behind setting the ideal maximum number for the overflow chip at half the modulus

src/chips/overflow/overflow_check.rs
sifnoc: Thanks for checking it!

> Also, just to be sure, the decomposition in performed little endian, namely the most significant bits is in the first advice column?

Yes, `a_1` is MSB in the description. also, fixed at [477c0cd](https://github.com/summa-dev/summa-solvency/pull/33/commits/477c0cdc7e81bef6d3d77c89a12a59bc2307a398)

src/merkle_sum_tree/utils/csv_parser.rs
sifnoc: Reverted at [141d71c](https://github.com/summa-dev/summa-solvency/pull/33/commits/141d71c376f760cfccee77d0a21a6260f77f1866)

enricobottazzi (approved):

sifnoc:
I plan to merge this after #29 is merged

alxkzmn (approved):

PR #43, opened 2023-06-09
Multiasset testing

### LINKED ISSUE #30: Update tests to account for multiple assets
- Generate the tree and hasher configs for the case of multiple assets (e.g., at least 2)
- Update test CSV data 
- Update tests

alxkzmn:
Modifies the circuit parameters and the tests for the case of two assets per user account. Resolves #30.


benches/full_solvency_flow.rs
```diff
@@ -69,7 +69,7 @@ fn generate_zk_proof_benchmark(_c: &mut Criterion) {
 
     let csv_file = format!("benches/csv/entry_2_{}.csv", LEVELS);
 
-    let assets_sum = [Fp::from(556863u64)]; // greater than liabilities sum (556862)
+    let assets_sum = [Fp::from(556863u64), Fp::from(556863u64)]; // greater than liabilities sum (556862)
```
enricobottazzi, line 72: Actually benchmarking doesn't work now because the csv files are based on a single balance. 
We can solve that by updating the csv files on AWS or by making the library more modular such that you can specify the N_ASSETS and all the related functions such as:

- csv_parser
- poseidon_params and specs

are modified accordingly.

src/merkle_sum_tree/utils/csv_parser.rs
```diff
@@ -5,12 +5,18 @@ use std::error::Error;
 use std::fs::File;
 use std::path::Path;
 
-#[derive(Debug, Deserialize)]
-struct CsvEntry {
-    username: String,
-    balance: Vec<String>,
+macro_rules! csv_entry {
```
enricobottazzi, line 8: The whole `csv_parser` doesn't seem very modular to me. I believe that if I want to use multi_assets 3, even if I generate the correct parameters, there's no way to parse it without manually modifying the csv_parser function.

src/merkle_sum_tree/utils/hash.rs
```diff
@@ -1,33 +1,34 @@
...
     hash_inputs[1..N_ASSETS + 1].copy_from_slice(&l2);
     hash_inputs[N_ASSETS + 1] = r1;
     hash_inputs[N_ASSETS + 2..2 * N_ASSETS + 2].copy_from_slice(&r2);
 
-    poseidon::Hash::<Fp, PoseidonSpecNode, ConstantLength<R_L_NODE>, WIDTH_NODE, R_L_NODE>::init()
+    poseidon::Hash::<Fp, PoseidonSpec, ConstantLength<L_NODE>, WIDTH, RATE>::init()
         .hash(hash_inputs)
 }
 
 pub fn poseidon_entry<const N_ASSETS: usize>(left: Fp, right: [Fp; N_ASSETS]) -> Fp {
```
enricobottazzi, line 26: I agree on this modification. 
Just for the record, before we were using 2 different Poseidon Hasher.
- 1 for hashing the entry which is an array `[username, balance#1, balance#2]` considering the case of 2 Assets
- 1 for hashing the node which is an arrat `[hashLeft, balance#1Left, balance#2Left, hashRight, balance#1Right, balance#2Right]`

The previous entry hasher was smaller in terms of RATE and WIDTH and therefore more efficient. But I think that this way makes it more simple.

src/circuits/tests.rs
```diff
@@ -67,12 +70,12 @@ mod test {
...
     #[test]
     fn test_valid_merkle_sum_tree_with_full_prover() {
-        let assets_sum = [Fp::from(556863u64)]; // greater than liabilities sum (556862)
+        let assets_sum = [Fp::from(556863u64), Fp::from(556863u64)]; // greater than liabilities sum (556862)
 
         let circuit = MerkleSumTreeCircuit::<LEVELS, MST_WIDTH, N_ASSETS>::init_empty();
 
         // we generate a universal trusted setup of our own for testing
-        let params = ParamsKZG::<Bn256>::setup(11, OsRng);
+        let params = ParamsKZG::<Bn256>::setup(12, OsRng);
```
enricobottazzi, line 78: I think we can safely revert it back to 11

src/circuits/tests.rs
```diff
@@ -427,7 +442,8 @@ mod test {
     // Passing an assets sum that is less than the liabilities sum should fail the lessThan constraint check
     #[test]
     fn test_is_not_less_than() {
```
enricobottazzi, line 444: Can you add a test to check the case in which the second `assets_sum` is not less_than while the first one is?

src/circuits/tests.rs
```diff
@@ -258,9 +261,9 @@ mod test {
     // Passing an invalid leaf balance as input for the witness generation should fail the permutation check between the computed root hash and the instance column root hash
```
enricobottazzi, line 261: I actually just realized that my previous test was designed wrongly as I was invalidating the leaf_hash rather than the leaf_balance. 

Now it makes more sense as there are 2 permutation checks that are failing:
- for the balance of the user for the first asset that doesn't match the expected one at row 1
- for the root hash that doesn't match the expected one passed at row 3 

Can you replace it to the comment of the test.

Also it would be cool to add another test where we invalidate the user_balance for the second asset. In that case I expect it to fail the on row 2 rather than row 1.

src/chips/poseidon/poseidon_spec.rs
```diff
@@ -1,23 +1,23 @@
...
 //! Patterned after [halo2_gadgets::poseidon::primitives::P128Pow5T3]
-use crate::chips::poseidon::poseidon_params_leaf;
+use crate::chips::poseidon::poseidon_params;
 use halo2_gadgets::poseidon::primitives::*;
 use halo2_proofs::arithmetic::Field;
 use halo2_proofs::halo2curves::bn256::Fr as Fp;
 
 #[derive(Debug, Clone, Copy)]
-pub struct MySpec;
+pub struct PoseidonSpec;
```
enricobottazzi, line 10: Even though we had this bug before, I think that the idea that Han suggested to just change the L while keeping the same specs for the Poseidon is still worth exploring. If we manage to use a single poseidon spec for any N_ASSETS that would simplify things a lot

src/merkle_sum_tree/utils/csv_parser.rs
alxkzmn: This would require a ; separator between "user" and "balances" which becomes a list of a length N_ASSETS.

sifnoc: Nice! I also thought about using the macro here, `csv_parser`.
I believe we should consider refactoring this after we've defined the API, as it closely ties in with API.
If everyone agrees, I'd be happy to create an issue for this refactoring task.

benches/full_solvency_flow.rs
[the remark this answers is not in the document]
enricobottazzi: @alxkzmn to open an issue for that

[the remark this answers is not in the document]
alxkzmn: https://github.com/summa-dev/summa-solvency/issues/46

src/merkle_sum_tree/utils/csv_parser.rs
alxkzmn: I believe this is already resolved by putting the array inside the 'balance' in CSV - this way the CSV should only have 'user' and 'balance' columns, but the records in 'balance' are expected to have N_ASSETS elements. Columns are semicolon-separated, and the balance values are comma-separated.

circuit_parameters_gen/generate_params.py
```diff
@@ -1,13 +1,19 @@
...
 # Parameters
 p = 0x30644E72E131A029B85045B68181585D2833E84879B9709143E1F593F0000001
 
-# First, set the parameters for the Merkle tree leaf hasher
+num_assets = int(sys.argv[1])  # Convert the command line argument to integer
+# Set the parameters for the Merkle tree leaf hasher
 # One hash and num_assets assets per Merkle tree node, plus 1 for the Poseidon "width" parameter
-num_assets = 1
-t = (1 + num_assets) + 1
+t = 7
```
enricobottazzi, line 16: Shouldn't `t` be a function of num_assets? Like it was before? Also t is the width or the rate?

src/circuits/tests.rs
```diff
@@ -427,11 +494,46 @@ mod test {
...
+                    offset: 0
+                },
+                cell_values: vec![
+                    // The zero means that is not less than
+                    (((Any::advice(), 49).into(), 0).into(), "0".to_string())
+                ]
+            }])
+        );
+
+        assert!(invalid_prover.verify().is_err());
```
enricobottazzi, line 529: I think this line is a left over, can be removed here and at line 553

enricobottazzi (changes requested):

src/chips/poseidon/poseidon_spec.rs
enricobottazzi: Ok, I'm not understanding it then. Why would you even need to generate different parameters if this is independent of the N_ASSETS

circuit_parameters_gen/generate_params.py
alxkzmn: No, it shouldn't. It is possible to use the same Poseidon config for variable hash input length. From the original paper: "Hashing multi-element objects with certain fields encoded as field elements, so that statements about these fields are proven in zero knowledge: We suggest variable-length sponge-based hashing with POSEIDON- 128 or POSEIDON-80 with width 5 (and rate 4)." Width of 5 didn't work for us because of that synthesis bug but width 7 works. Meanwhile, we are already using this single config with two different input lengths for Entry and Node.

alxkzmn: 't' is the width.

src/chips/poseidon/poseidon_spec.rs
alxkzmn: Removed the reference to the number of assets.

src/merkle_sum_tree/utils/csv_parser.rs
sifnoc: I understand, it's a very straightforward solution! I like it.

sifnoc (approved):

Issue #7, opened 2023-04-13
Add on-chain EVM verifier


enricobottazzi:
The feature relates to adding the possibility to add an EVM verifier to the proving system.

More on that on the Masterplan

In order to deploy a on-chain evm verifier these library might be used:

- [Snark Verifier PSE](https://github.com/privacy-scaling-explorations/snark-verifier)
- [EZKL](https://github.com/zkonduit/ezkl#verifying-with-the-evm-) to support transforming the bytecode into a deployable smart contract and => https://github.com/zkonduit/ezkl/blob/c6aed1499b541c9310ee55516e64879b089aaaac/src/execute.rs#L451

PR #48, opened 2023-06-15
Feat recursive verifier

enricobottazzi:
Issue #7 

Write up here => https://hackmd.io/bCCfhHRPS5S4zzNCpuWAIw

### DIFF
--- a/zk_prover/src/circuits/tests.rs
+++ b/zk_prover/src/circuits/tests.rs
@@ -1,18 +1,27 @@
 #[cfg(test)]
 mod test {
 
-    use crate::circuits::merkle_sum_tree::MerkleSumTreeCircuit;
-    use crate::circuits::utils::{full_prover, full_verifier, generate_setup_params};
-    use crate::merkle_sum_tree::big_int_to_fp;
-    use crate::merkle_sum_tree::{MST_WIDTH, N_ASSETS};
+    use crate::circuits::{
+        aggregation::WrappedAggregationCircuit,
+        merkle_sum_tree::MerkleSumTreeCircuit,
+        utils::{full_prover, full_verifier, generate_setup_params},
+    };
+    use crate::merkle_sum_tree::{big_int_to_fp, MST_WIDTH, N_ASSETS};
+    use ark_std::{end_timer, start_timer};
     use halo2_proofs::{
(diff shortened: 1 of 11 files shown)


alxkzmn (approved):

zk_prover/src/circuits/tests.rs
```diff
@@ -104,6 +113,131 @@ mod test {
         assert!(full_verifier(&params, &vk, proof, &public_input));
     }
 
+    #[test]
+    #[ignore]
+    fn test_valid_merkle_sum_tree_with_full_recursive_prover() {
+        // params for the aggregation circuit
+        let params_agg = generate_setup_params(23);
```
sifnoc, line 120: Could you provide some insight on why we need such a large number of rows here?

enricobottazzi: Sure, I'm quoting Han here =>

The k of aggregation circuit would be roughly `log2(30000*n)` where n is
`n = #columns + #internal_columns + #quotient_commitments + #pc_opening_commitments + 2*#old_accumulators`
where
- #columns are allocated by user e.g. meta.xxx_column() or meta.selector()
- #internal_columns are used by lookup argument and permutation argument
- #pc_opening_commitment depends on which kind of polynomial commitment scheme we are using

So in short it is mainly only about the #columns of the input circuit, so more columns -> larger aggregation circuit -> longer aggregation time. 

I think we can potentially move it down to 22, but not much more that this. As you can notice [here](https://github.com/privacy-scaling-explorations/snark-verifier/blob/main/snark-verifier-sdk/examples/standard_plonk.rs#L162) also the simple plonk example provided in the SDK requires k to be 22

sifnoc (approved):
Fantastic! This is another scalable solution for us!

zk_prover/src/circuits/tests.rs
enricobottazzi: Hi @sifnoc, for `generate_setup_params` I suggest you downloading the relative `hermez_raw` file from https://github.com/han0110/halo2-kzg-srs. 

In the `README` it is described how to use that => https://github.com/summa-dev/summa-solvency/tree/master/zk_prover#powers-of-tau-trusted-setup

Also, in the latest optimization the `k` of the aggregation circuit will likely to be reduced to 21 or 20

PR #50, opened 2023-06-15
Minor changes!

enricobottazzi:
As the title says, I applied a few minor changes in this PR.

I suggest to follow along the title of the commits and the related description to understand them. 

The core idea is just to polish the code and save lines when possible


sifnoc (approved):
Much more clear code lines! Thanks

Issue #49, opened 2023-06-15
Remove balance from public input


enricobottazzi:
The balance was added to public input as a way to test the circuit. 

In production, the balance of a user should not be public, as it is committed within the `leaf_hash`.

PR #51, opened 2023-06-15
remove balance from public input

enricobottazzi:
Issue #49 

What is affected by this fix?

- The instance column of the circuit should no longer contain the `balances` 
- The `synthesize` of the circuit should no longer call `chip.expose_public` for the `leaf_balances`
- All the other `expose_public` functions within `synthesize` of the circuit should adjust for the new row mapping inside the instance column
- In the `MerkleSumTreeChip` the function `enforce_less_than` copies the assets_sum from instance. It has to be modified to copy assets at different rows.
- All the tests need to be adjusted accordingly

| Instance  |
| --------- |
| leaf_hash |
| leaf_balance_asset_0          |
| leaf_balance_asset_1          |
| root_hash          |
| assets_sum_0          |
| assets_sum_1          |

| Instance after PR |
| --------- |
| leaf_hash |
| root_hash          |
| assets_sum_0          |
| assets_sum_1          |

On top of that, a further fixing has been applied in c8a9fca => in `assign_leaf_hash_and_balances` the `leaf_hash` gets assigned from the instance column rather assigning it from witness generating and later comparing it to the `leaf_hash`. This reduces a copy constraint in the circuit.

Note that because of this modification the whole test `test_invalid_leaf_hash_as_witness` has been removed as there's no longer such thing as passing an invalid leaf hash as witness 

EDIT: I reverted the changes in c8a9fca. As part of another PR I actually realized that the API was more helpful if working like this

### DIFF
--- a/zk_prover/benches/full_solvency_flow.rs
+++ b/zk_prover/benches/full_solvency_flow.rs
@@ -4,6 +4,7 @@ use halo2_proofs::{
     plonk::{keygen_pk, keygen_vk},
     poly::kzg::commitment::ParamsKZG,
 };
+use snark_verifier_sdk::CircuitExt;
 use summa_solvency::{
     circuits::merkle_sum_tree::MerkleSumTreeCircuit,
     circuits::utils::{full_prover, full_verifier, generate_setup_params},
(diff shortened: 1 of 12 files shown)


alxkzmn (approved):

sifnoc (approved):
Good! I create an issue related to this.
Verification 'leaf_hash' https://github.com/summa-dev/summa-solvency/issues/64

Issue #53, opened 2023-06-16
Separate Logic of Proof Of Inclusion and Proof Of Solvency in 2 different circuits


enricobottazzi:
Currently, the `MerkleSumTreeCircuit` performs two type of checks leveraging the `MerkleSumTreeChip`: 

- checks that the user identified by the `leaf_hash` is included in the merkle sum tree identified by the public commitment `root_hash`. We define it **proof of inclusion**.
- checks that the `computed_sum` is less than the `assets_sum`. We define it **proof of solvency**.

According to this implementation, every time that a proof is generated by the CEX for a user, this will both include a proof of inclusion and a proof of solvency. 

But this approach is highly inefficient. The CEX can generate a proof of solvency only once and provide individual proof of inclusion to each user.

The proof of solvency circuit would follow this logic:

```
    // check 1: Assets greater than liabilities
    ΣcexAssetsSum >= rootBalanceLeft + rootBalanceRight
    
    // check 2: Hash of the last level of the tree should match the publicly committed root hash
    root_hash === H(rootHashLeft, rootBalanceLeft, rootHashRight, rootBalanceRight)
```

The result of the PR will be two new circuits `mst_inclusion` and `solvency`. Both of them can rely on the methods provided by the `MerkleSumTreeChip`

PR #60, opened 2023-06-19
Feat 53

enricobottazzi:
solves #53


enricobottazzi:
There's still an issue here: Both the circuits are using the same configuration, resulting in a lot of unused columns.

PR #68, opened 2023-06-26
Overflow chip with reduced column usage

sifnoc:
Modified overflow chip to utilize only two columns, one for `value` and the other one for `decomposed_values`.


zk_prover/src/chips/overflow/overflow_check.rs
```diff
@@ -6,43 +6,48 @@ use halo2_proofs::plonk::{Advice, Column, ConstraintSystem, Error, Expression, F
...
+        meta: &mut ConstraintSystem<Fp>,
+        a: Column<Advice>,
+        b: Column<Advice>,
+    ) -> OverflowCheckConfig<MAX_BITS> {
         let range = meta.fixed_column();
-        let a = meta.advice_column();
-        let decomposed_values = [(); ACC_COLS].map(|_| meta.advice_column());
+        let toggle_overflow_check = meta.complex_selector();
+
+        let num_rows = MOD_BITS / MAX_BITS as usize;
```
enricobottazzi, line 37: It may end up that num_rows is not an integer number if MAX_BITS is an odd number. I don't know if there's a situation in which makes sense to use odd `MAX_BITS`, maybe we can add an assert to avoid this.

zk_prover/src/chips/overflow/overflow_check.rs
```diff
@@ -6,43 +6,48 @@ use halo2_proofs::plonk::{Advice, Column, ConstraintSystem, Error, Expression, F
 use halo2_proofs::poly::Rotation;
 use std::fmt::Debug;
 
+const MOD_BITS: usize = 252;
```
enricobottazzi, line 9: I suggest adding this as generic to the Config too, so one can modify the range check freely

enricobottazzi (changes requested):
The PR LGTM after these small changes. After #67 gets merged, I'd like to explore other ways of designing this that may lead to more efficiency. For example spreading the values across the 4 advice columns that we have at our disposal, in order to reduce the number of rows used and the relative offset!

alxkzmn (approved):

zk_prover/src/chips/overflow/overflow_check.rs
sifnoc: The num_rows is currently calculated as an integer, regardless of whether `MAX_BITS` is an odd number. |
However, I think it's better to add a warning sign for when `MAX_BITS` is an odd number, as future developers might expect different results.

zk_prover/src/chips/overflow/overflow_check.rs
sifnoc: Ah, it makes sense. I will fix and revert 16bits test cases.

sifnoc: Fixed at [7da5277](https://github.com/summa-dev/summa-solvency/pull/68/commits/7da5277dac72057e94354e80442d60ee1b63bb7a)

zk_prover/src/chips/overflow/overflow_check.rs
sifnoc: Added Warn at [ffce080](https://github.com/summa-dev/summa-solvency/pull/68/commits/ffce0801a8248d428b00d7e0c954c8a2ee794fc4)

enricobottazzi (approved):
🚀

zk_prover/src/chips/overflow/overflow_check.rs
sifnoc: this commit checks whether the `MOD_BITS` was divided evenly by `MAX_BITS`

PR #76, opened 2023-06-30
optimize: lookup check performs no relative rotations

### LINKED ISSUE #74: Improvement on `OverflowChip` efficiency
Currently, the overflow chip makes a big set of rotations when performing the lookup which increases proving time and proof size (and indirectly verification cost)

https://github.com/summa-dev/summa-solvency/blob/eee534af6311d89b9c91c36d27f1542bbffa5909/zk_prover/src/chips/overflow/overflow_check.rs#L101-L107

Remember that the rotations are always relative to the selector that enforces the constraint. Here's where the selector gets toggled

https://github.com/summa-dev/summa-solvency/blob/master/zk_prover/src/chips/overflow/overflow_check.rs#L127

In order to increase efficiency, it is possible to just add a new selector for the lookup check that gets activated using the rows of advice column `b` as offset and queries `cell` and `range` at the current rotation

```
         let cell = meta.query_advice(b, Rotation::cur()); 
         let range = meta.query_fixed(range, Rotation::cur()); 
```

The downside of this modification is that it adds a further selector column

enricobottazzi:
Fixes #74 

To do:

- [x] Wait for #75 to be merged and run new benches


alxkzmn (approved):

zk_prover/src/chips/overflow/overflow_check.rs
```diff
@@ -98,20 +100,23 @@ impl<const MAX_BITS: u8, const MOD_BITS: usize> OverflowChip<MAX_BITS, MOD_BITS>
...
-                let enable_lookup = meta.query_selector(toggle_overflow_check);
-                vec![(enable_lookup * cell, range)]
-            });
-        }
+        meta.lookup_any("range check for MAXBITS", |meta| {
+            let cell = meta.query_advice(b, Rotation::cur());
+            let range = meta.query_fixed(range, Rotation::cur());
+
+            // print range
+            println!("range: {:?}", range);
```
sifnoc, line 108: Did this line of code possibly remain here by mistake after testing?

enricobottazzi: Yep 🥲 fixed at f029de

PR #78, opened 2023-06-30
LTChip - vertical approach

ameya-deshmukh:
Implemented a vertical approach for ```LTChip``` to reduce the number of advice columns.


zk_prover/src/chips/less_than_vertical.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Chip that compares lhs < rhs.
+#[derive(Clone, Debug)]
+pub struct LtVerticalChip<const N_BYTES: usize> {
+    config: LtVerticalConfig<N_BYTES>,
+}
+
+impl<const N_BYTES: usize> LtVerticalChip<N_BYTES> {
+    /// Configures the LtVertical chip.
+    pub fn configure(
```
enricobottazzi, line 59: Can you take `lt`, `diff` and `u8` as input of the `configure` function? In general we prefer to declare all the columns in the `syntesize` function of the circuit and downstream it to the chips such that we can share columns across chips

zk_prover/src/chips/less_than_vertical.rs
```diff
@@ -0,0 +1,204 @@
...
+            let q_enable = q_enable(meta);
+            let lt = meta.query_advice(lt, Rotation::cur());
+
+            let mut diff_bytes = Vec::<Expression<Fp>>::new();
+
+            diff_bytes.push(meta.query_advice(diff, Rotation::cur()));
+            diff_bytes.push(meta.query_advice(diff, Rotation::next()));
+
+            for i in 2..N_BYTES {
+                diff_bytes.push(meta.query_advice(diff, Rotation(i as i32)));
```
enricobottazzi, line 80: I believe that this can be as a for loop from 0 to N_Bytes. Rotation(0) should be equivalent to `Rotation::cur()`.

zk_prover/src/chips/less_than_vertical.rs
```diff
@@ -0,0 +1,204 @@
...
+            vec![(u8_cell, u8_range)]
+        });
+
+        meta.lookup_any("range check for u8", |meta| {
+            let u8_cell = meta.query_advice(diff, Rotation::next());
+            let u8_range = meta.query_fixed(u8, Rotation::cur());
+            vec![(u8_cell, u8_range)]
+        });
+
+        for i in 0..N_BYTES {
```
enricobottazzi, line 107: Considered the previous comment, I think loop should work fine as it is and you don't need the two previous `meta.lookup_any` with `cur` and `next` rotations.

enricobottazzi (changes requested):

enricobottazzi:
@ameya-deshmukh LGTM, can you add the unit tests? The ones from the original zk-evm gadget should be enough

ameya-deshmukh:
Sure sir, adding them.

ameya-deshmukh:
Yeah sure. I'll have a look.

ameya-deshmukh (commented):
@enricobottazzi thanks for this! Sorry for overlooking and not omitting the unnecessary ```Fp``` in ```TestCircuit``` since we're not using generics anymore. Also, curious why you removed the looping for lookup. Since we're following a row-based and not a column based approach, shouldn't the lookup loop over all the rows? Or am I missing something here?

enricobottazzi:
@ameya-deshmukh, I fixed the bugs in testing. 

- The lookup was not being performed correctly. Now it is toggled by the `q_enable` 
- With the vertical implementation we could not test a lot of tuples `(lhs, rhs)` in the same region because they would call the assign function at any `idx + 1` offset and this would basically rewrite to the `diff` column. 

For example, if we only add a single tuple with `assign` called at offset 0, the witness table would look like this (so everything good):

| value[a] | value[b] | diff |
| -------- | -------- | ---- |
| x             |   z      |0xff  |
|               |          |0xff  |
|               |          |0xff  |
|              |          | 0xff |


But if we add another tuple and assign it at offset 1, it will rewrite diff at offset 1, which is clearly not an intended behavior

| value[a] | value[b] | diff |
| -------- | -------- | ---- |
| x        |   z      |0xff  |
| y        |   n      |0xdd  |
|          |          |0xdd  |
|          |          | 0xdd |
|          |          | 0xdd |

For the way we use the chip in our `SolvencyCircuit` this fixed implementation should be fine

enricobottazzi (approved):

ameya-deshmukh:
Ah got it, thanks! :)

sifnoc (approved):

zk_prover/src/chips/less_than_vertical.rs
```diff
@@ -0,0 +1,531 @@
...
+        &self.config
+    }
+
+    fn loaded(&self) -> &Self::Loaded {
+        &()
+    }
+}
+
+#[cfg(test)]
+mod test {
```
alxkzmn, line 200: Let's have the tests in a separate file, as we do with other code in the project.

alxkzmn (changes requested):

zk_prover/src/chips/less_than_vertical.rs
enricobottazzi: Good point! I'll take it

alxkzmn (approved):

PR #79, opened 2023-07-03
Implement Summa smart contract

alxkzmn:


contracts/src/Summa.sol
```diff
@@ -0,0 +1,135 @@
...
+    ) external view returns (bool);
+}
+
+contract Summa is Ownable {
+    using ECDSA for bytes32;
+
+    uint256 constant ETH_INPUTS = 1;
+    uint256 constant ERC20_INPUTS = 1;
+
+    struct PublicInput {
```
enricobottazzi, line 24: Can we separate the public input of the circuit from the other values that participate to the Proof Of Solvency but are not actual input of the circuit? Such as the addresses and the signatures.

contracts/src/Summa.sol
```diff
@@ -0,0 +1,135 @@
...
+    function verify(
+        uint256[] calldata pubInputs,
+        bytes calldata proof
+    ) external view returns (bool);
+}
+
+contract Summa is Ownable {
+    using ECDSA for bytes32;
+
+    uint256 constant ETH_INPUTS = 1;
```
enricobottazzi, line 21: Is there a way to not hardcode ETH_INPUTS and ERC20_INPUTS?

contracts/src/Summa.sol
enricobottazzi, line 24: Can we separate the actual public inputs of the circuits from the variable that participate to the proof of solvency but are not strictly public inputs of the circuit? Such as the addresses and the signatures

enricobottazzi (changes requested):

contracts/src/Summa.sol
alxkzmn: If we don't hard-code them, we'll have to hard-code the input balance array size. The verifier always expects an input array of a known size so I don't think it's a problem to have that number explicitly defined.

enricobottazzi: I see now! Actually the lenght of the array `inputs` that is taken by the verifier is not depended on `ETH_INPUTS` and `ERC20_INPUTS`. 

For example, let's say that a CEX has 20 wallets with eth on them. The wallet provides all length-20-arrays `cexETHAddresses`, `cexETHBalances` and `cexEthAddressSignatures` to the `submitProofOfSolvency` function. The public input of the circuit will be `cexETHAssetsSum` which must be equal to ƩcexETHBalances.

contracts/src/Summa.sol
```diff
@@ -0,0 +1,131 @@
...
+    function submitProofOfSolvency(
+        string memory exchangeId,
+        address[] memory cexETHAddresses,
+        address[] memory cexERC20Addresses,
+        uint256[] memory cexETHBalances,
+        uint256[] memory cexERC20Balances,
+        address[] memory erc20ContractAddresses,
+        bytes[] memory cexEthAddressSignatures,
+        bytes[] memory cexERC20AddressSignatures,
+        uint256 mstRoot,
```
sifnoc, line 41: Could you implement a feature to simplify the retrieval of `mstRoot`? 
I believe it would be very convenient to have a mapping like this: `mapping(bytes32 => uint256) public exchangeIdToMstRoot`. 
This would greatly contribute to ensuring reliability during user-side verification.

alxkzmn: Sure, will do!

contracts/src/Summa.sol
```diff
@@ -0,0 +1,116 @@
...
+    function verify(
+        uint256[] calldata pubInputs,
+        bytes calldata proof
+    ) external view returns (bool);
+}
+
+contract Summa is Ownable {
+    using ECDSA for bytes32;
+
+    uint256 constant ERC20_INPUTS = 1;
```
enricobottazzi, line 21: I think this is a leftover

contracts/src/Summa.sol
alxkzmn: I gave it some thought and ended up including the MST root in the emitted event. Normally, you would want to write to the contract state only in the following cases:

- the state needs to be accessible by the contract for its future computations (e.g., in a DeFi protocol you need to store the modified state in the contract, but in our case we don't perform any computations on previous data);
- the state needs to be accessible by other contracts (again, think of DeFi composability, and we don't need any composability yet).

Events are very convenient for us because the user would probably want to verify a proof at a certain point in time that corresponds to a certain block timestamp, which is exactly how you query the contract events using filters:

```
 const filter = summaContract.filters.ProofOfSolvencySubmitted(exchangeId);
 const fetchEvents = async () => {
     const events = await summaContract.queryFilter(filter, startingBlock, endingBlock);
     ...
 };
```
It's a common practice to offload the data that's already recorded on chain (i.e., you can always find the TX and read the mstRoot function argument value) in a transaction log to save on gas.

enricobottazzi (changes requested):
Everything LGTM, just a minor thing before approving

contracts/src/Summa.sol
sifnoc: Yes! It seems like a better way to get info without additional costs! Thanks

enricobottazzi (approved):

sifnoc (approved):

PR #80, opened 2023-07-03
feat: add solidity verifier

enricobottazzi:
In this PR: 

- Add APIs in `utils` to generate a solidity verifier starting from a circuit and to generate the solidity calldata to submit such proof on chain for verification 
- Reduced poseidon `WIDTH` to 1 and `RANGE_BITS` to 64 to make on-chain verification efficient
- Added test to deploy on-chain verifier for `SolvencyCircuit` and verify such proof on chain. 
- Added proving key for both `SolvencyCircuit` and `MstInclusionCircuit` to backend artifcats


zk_prover/src/circuits/tests.rs
```diff
@@ -76,6 +77,34 @@ mod test {
         assert!(full_verifier(&params, &vk, proof, circuit.instances()));
     }
 
+    #[test]
+    fn test_valid_solvency_with_full_prover() {
+        let circuit = SolvencyCircuit::<L, N_ASSETS, N_BYTES>::init_empty();
+
+        // we generate a universal trusted setup of our own for testing
+        let params = generate_setup_params(10);
```
sifnoc, line 85: Is the K for the Solvency circuit always have a value that is one less than the K for the Inclusion circuit?

zk_prover/src/circuits/tests.rs
```diff
@@ -126,6 +157,12 @@ mod test {
             num_instances,
             None,
         );
+
+        println!(
```
sifnoc, line 161: Is this a leftover code?

zk_prover/src/circuits/tests.rs
```diff
@@ -501,6 +538,55 @@ mod test {
...
+        let pk = gen_pk(&params, &circuit, None);
+
+        get_verification_cost(&params, &pk, circuit.clone());
+
+        let num_instances = circuit.num_instance();
+        let instances = circuit.instances();
+
+        let proof_calldata = gen_evm_proof_shplonk(&params, &pk, circuit, instances.clone());
+
+        // let yul_code_path = "../contracts/src/SolvencyVerifier.yul";
```
sifnoc, line 561: It seems there might be some leftovers in this test.

zk_prover/src/merkle_sum_tree/tests.rs
```diff
@@ -95,39 +95,39 @@ mod test {
         }
     }
 
-    #[test]
-    fn test_mst_with_bigint() {
-        // create new merkle tree with entries that have balances greater than 2^64
-        let merkle_tree =
-            MerkleSumTree::<N_ASSETS>::new("src/merkle_sum_tree/csv/entry_16_bigints.csv").unwrap();
+    // #[test]
```
sifnoc, line 98: Do you plan to use this test later?

zk_prover/src/merkle_sum_tree/params.rs
```diff
@@ -7,4 +7,4 @@ pub const L_NODE: usize = 2 * (1 + N_ASSETS);
 /// Poseidon hasher parameter for Length used in MST entries (aka levaes, they only take one hash and one set of assets as input)
 pub const L_ENTRY: usize = 1 + N_ASSETS;
 /// Overflow limit for the accumulated balances in the proof of solvency
-pub const MOD_BITS: usize = 248;
+pub const RANGE_BITS: usize = 64;
```
sifnoc, line 10: Would you mind changing this word inside the overflow chip too?

zk_prover/README.md
```diff
@@ -62,22 +62,28 @@ Furthermore the benchmarking function `verify_zk_proof_benchmark` will also prin
 
 Run on MacBook Pro 2023, M2 Pro, 32GB RAM, 12 cores
 
+Benches run after PR #80 (`add solidity verifier`). In order to achieve small proof size, to be cheap to verify on-chain. We decreased the RANGE_BITS in which the balances should lie to 64 and reduced the rate of the Poseidon hash function to 1.
```
alxkzmn, line 65: You mention this same pull request here, and it already has the benches - is that correct?

alxkzmn (changes requested):

zk_prover/src/circuits/tests.rs
enricobottazzi: Yes, I removed this too in 9b73cd0. in Just pointing it to @alxkzmn too as a way to show how to use the API to write the solidity verifier contract and get the solidity calldata

zk_prover/src/circuits/tests.rs
enricobottazzi: Removed in 9b73cd0.

zk_prover/src/merkle_sum_tree/tests.rs
enricobottazzi: Thanks for pointing it out. Removed it in 9b73cd0. 
I don't think this is crucial to maintain as test

zk_prover/src/merkle_sum_tree/params.rs
enricobottazzi: yes! fixed in 0d38775

zk_prover/README.md
enricobottazzi: Yes, these benches are run based on this branch. Is that ok?

Also, I am aware that the process of benchmarking should be more structured. Do you have any experience with that?

enricobottazzi: Yes, these benches are run based on this branch. Is that ok?

Also, I am aware that the process of benchmarking should be more structured. Do you have any experience with that?

sifnoc (approved):
Let's merge them!

zk_prover/README.md
alxkzmn: I mean do you really want to leave a reference to some PR in the docs?

alxkzmn (approved):

Issue #82, opened 2023-07-04
Implement SnapShotData 


sifnoc:
More detail doc in here https://hackmd.io/0Dp2mudRRCKPov6jI8LYhg?view#SnapShotData

This component role can be like these

[ ] Aggregating User Data, Entry
[ ] Generating Proofs
[ ] Serving Signed Account Data & Proof(s)

PR #88, opened 2023-07-05
SnapshotData in backend

sifnoc:
This is related to task #82.

This PR contains the implementation of the `SnapshotData` struct along with its associated methods. 
The main purpose of `SnapshotData` is to streamline the handling and processing of assets and entries.

The PR is divided into two parts:

- Parsing CSV Data for assets
This component introduces a function named `parse_csv_to_assets`, which is designed to parse asset information from a CSV file and store it within the `SnapshotData` struct. 
It reads the CSV data, processes it into Asset structs (including public keys, balances, and signatures), and calculates the total balance for each asset type.

- SnapshotData Struct
This struct incorporates fields for storing asset data, proofs for individual users, aggregated balances, and on-chain proof. 
It is an essential part of managing and processing the data.

You can run the test with a 'testing' feature flag to skip generating the recursive proof. 

Use the following command:
`> cargo test --features testing."`


enricobottazzi (changes requested):

backend/src/apis/csv_parser.rs
```diff
@@ -0,0 +1,94 @@
+use crate::apis::snapshot_data::Asset;
+use halo2_proofs::halo2curves::bn256::Fr as Fp;
+use halo2_proofs::halo2curves::ff::PrimeField;
+use halo2_proofs::halo2curves::serde::SerdeObject;
+// use halo2_proofs::halo2curves::serde::SerdeObject;
+use num_bigint::BigInt;
```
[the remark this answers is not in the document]
sifnoc: Fixed at [7c52421](https://github.com/summa-dev/summa-solvency/pull/88/commits/7c52421c2d9141edcdcb123cd03cec39bf56d3c8)

backend/src/apis/csv/assets_2.csv
```diff
@@ -0,0 +1,4 @@
+name;pubkey;balances;signature
+eth;0x627306090abaB3A6e1400e9345bC60c78a8BEf57;1500;3045022100c12a7d54972f26d4f4766b8ad5d7a3d7cfe1b5ce3c1e1a9116a6a25db11d8d7a0220476fc1d66f673f52f9dd8c637f62f6e4e2a31dbca5f65fddbca9f3faff9f6d6b
```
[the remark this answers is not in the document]
sifnoc: currently the balances are solely utilized for computing the 'assets_sum'. 

> I guess that an idea for the future would be to fetch the balances for each pubkey from some api service.

Yes, we have to think about it

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+    },
+    merkle_sum_tree::{Entry, MerkleSumTree},
+};
+
+const LEVELS: usize = 4;
+const L: usize = 2 + (N_ASSETS * 2);
+const K: u32 = 11;
+const N_BYTES: usize = 31; // 248 / 8
+const N_ASSETS: usize = 2;
+
```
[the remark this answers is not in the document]
sifnoc: Yes! I think it's more sense,
Update at 7c52421c2d9141edcdcb123cd03cec39bf56d3c8

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+#[derive(Debug)]
+struct SnapshotData<const N_ASSETS: usize> {
+    exchange_id: String,
+    commit_hash: Fp,                          // May change type later
```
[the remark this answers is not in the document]
sifnoc: Fixed at 7c52421c2d9141edcdcb123cd03cec39bf56d3c8

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+    entries: HashMap<usize, Entry<N_ASSETS>>, // TODO: change to `HashMap<Name, Entry<N_ASSETS>>`
+    assets: Vec<Asset>,
+    user_proofs: Option<HashMap<Name, InclusionProof<N_ASSETS>>>,
+    on_chain_proof: Option<Vec<u8>>,
+}
```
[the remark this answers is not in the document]
sifnoc: Sure!
Update at [062ef7d](https://github.com/summa-dev/summa-solvency/commit/062ef7d80254b3f82cc34d1a768c205cb0f726c8)

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+const N_BYTES: usize = 31; // 248 / 8
+const N_ASSETS: usize = 2;
+
+#[derive(Debug)]
+struct SnapshotData<const N_ASSETS: usize> {
+    exchange_id: String,
+    commit_hash: Fp,                          // May change type later
+    entries: HashMap<usize, Entry<N_ASSETS>>, // TODO: change to `HashMap<Name, Entry<N_ASSETS>>`
+    assets: Vec<Asset>,
+    user_proofs: Option<HashMap<Name, InclusionProof<N_ASSETS>>>,
```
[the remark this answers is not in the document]
sifnoc: Indeed! Storing all proofs is not practical; this is just a temporary data type. 
We need to look for a more efficient data structure for the final version. 

By the way, the 'entries' and 'commit_hash' fields are now replaced with 'MerkleSumTree'.
At [e9cd091](https://github.com/summa-dev/summa-solvency/pull/88/commits/e9cd091d9f7ffee32627a09b0eb9490ba6f4e6ac)

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+const L: usize = 2 + (N_ASSETS * 2);
+const K: u32 = 11;
+const N_BYTES: usize = 31; // 248 / 8
+const N_ASSETS: usize = 2;
+
+#[derive(Debug)]
+struct SnapshotData<const N_ASSETS: usize> {
+    exchange_id: String,
+    commit_hash: Fp,                          // May change type later
+    entries: HashMap<usize, Entry<N_ASSETS>>, // TODO: change to `HashMap<Name, Entry<N_ASSETS>>`
```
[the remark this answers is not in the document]
sifnoc: Yup, good point. also updated at [e9cd091](https://github.com/summa-dev/summa-solvency/pull/88/commits/e9cd091d9f7ffee32627a09b0eb9490ba6f4e6ac)

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct Asset {
+    pub name: String,
+    pub pubkeys: Vec<String>,
+    pub balances: Vec<BigInt>,
+    pub sum_balances: Fp,
+    pub signature: Vec<String>,
+}
+
+#[derive(Debug)]
+struct InclusionProof<const N_ASSETS: usize> {
```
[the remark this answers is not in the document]
sifnoc: Ah, indeed, I made this base on the previous contract code.
Fixed At 062ef7d

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+        let vk = keygen_vk(&params, &circuit).expect("vk generation should not fail");
+        let pk = keygen_pk(&params, vk.clone(), &circuit).expect("pk generation should not fail");
+
+        // Only now we can instantiate the circuit with the actual inputs
+        let solvency_circuit = SolvencyCircuit::<L, N_ASSETS, N_BYTES>::init(entry_csv, assets_sum);
+
+        return (solvency_circuit, pk);
+    }
+
+    fn generate_agg_circuit(
```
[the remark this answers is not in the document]
sifnoc: I see, I thought the `submitProofOfSolvency` needs proof from the `aggregation circuit`. 
Fixed and replaced 062ef7d80254b3f82cc34d1a768c205cb0f726c8

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+const N_ASSETS: usize = 2;
+
+#[derive(Debug)]
+struct SnapshotData<const N_ASSETS: usize> {
+    exchange_id: String,
+    commit_hash: Fp,                          // May change type later
+    entries: HashMap<usize, Entry<N_ASSETS>>, // TODO: change to `HashMap<Name, Entry<N_ASSETS>>`
+    assets: Vec<Asset>,
+    user_proofs: Option<HashMap<Name, InclusionProof<N_ASSETS>>>,
+    on_chain_proof: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
sifnoc: Sure! Fixed at f81e50e

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn get_onchain_proof(&self) -> Result<Vec<u8>, &'static str> {
+        match &self.on_chain_proof {
+            Some(proof) => Ok(proof.clone()),
+            None => Err("on-chain proof not initialized"),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
sifnoc: Deleted at f81e50e
We can recover the test file later for testing

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,297 @@
...
+        // Check assets
+        assert!(snapshot_data.assets[0].name.contains(&"eth".to_string()));
+        assert!(snapshot_data.assets[1].name.contains(&"dai".to_string()));
+        assert!(snapshot_data.assets[0].balances[0] == BigInt::from(1500u32));
+        assert!(snapshot_data.assets[0].balances[1] == BigInt::from(2500u32));
+    }
+
+    #[test]
+    fn test_snapshot_data_generate_proof() {
+        let entry_csv = "src/apis/csv/entry_16.csv";
```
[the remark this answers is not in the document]
sifnoc: Fixed at [e9cd091](https://github.com/summa-dev/summa-solvency/pull/88/commits/e9cd091d9f7ffee32627a09b0eb9490ba6f4e6ac)

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+    proof_of_solvency: Option<SolvencyProof<N_ASSETS>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct Asset {
+    pub name: String,
+    pub pubkeys: Vec<String>,
+    pub balances: Vec<BigInt>,
+    pub sum_balances: Fp,
+    pub signature: Vec<String>,
```
enricobottazzi, line 42: replace with `signatures` ?

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+    const L: usize,
+    const N_ASSETS: usize,
+    const N_BYTES: usize,
+    const K: u32,
+> {
+    exchange_id: String,
+    mst: MerkleSumTree<N_ASSETS>,
+    assets: Vec<Asset>,
+    proofs_of_inclusion: HashMap<u64, InclusionProof>,
+    proof_of_solvency: Option<SolvencyProof<N_ASSETS>>,
```
enricobottazzi, line 33: Why this is an option and `proofs_of_inclusions` are not?

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub name: String,
+    pub pubkeys: Vec<String>,
+    pub balances: Vec<BigInt>,
+    pub sum_balances: Fp,
+    pub signature: Vec<String>,
+}
+
+#[derive(Debug, Clone)]
+struct InclusionProof {
+    // public input
```
enricobottazzi, line 47: The public inputs of InclusionProof are `leaf_hash` and `root_hash`

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+    leaf_hash: Fp,
+    vk: Vec<u8>,
+    proof: Vec<u8>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
```
enricobottazzi, line 53: What is `PartialEq` ?

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+struct SolvencyProof<const N_ASSETS: usize> {
```
enricobottazzi, line 54: The public inputs of `SolvencyProof` are :

- `root_hash`
- `assets_sum`

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+            mst,
+            assets,
+            proofs_of_inclusion: user_proofs,
+            proof_of_solvency: None,
+        })
+    }
+
+    fn generate_inclusion_proof(&self, user_index: usize) -> Result<InclusionProof, &'static str> {
+        // TODO: fetch pk from outside. For now, we generate them here
+        let circuit = MstInclusionCircuit::<LEVELS, L, N_ASSETS>::init_empty();
```
enricobottazzi, line 91: I think it's ok to generate the params locally. The problem is that this way we are creating a new random `params` for each `inclusion_proof` which will then result in different `pk` and `vk`. So I think that we can safely generate the params locally, but we 
need to do it only once, maybe during `init()` and store the reuslting `pk` and `vk` somewhere in `Snapshot` before any proof generation function gets called

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+    vk: Vec<u8>,
+    proof: Vec<u8>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct SolvencyProof<const N_ASSETS: usize> {
+    // public inputs
+    penultimate_node_hash: [Fp; 2],
+    assets_sum: [Fp; N_ASSETS],
+    vk: Vec<u8>,
```
enricobottazzi, line 58: We don't need to store the vk here, since it gets verified inside the smart contract, the smart contract already has the verification key encoded.

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub balances: Vec<BigInt>,
+    pub sum_balances: Fp,
+    pub signature: Vec<String>,
+}
+
+#[derive(Debug, Clone)]
+struct InclusionProof {
+    // public input
+    leaf_hash: Fp,
+    vk: Vec<u8>,
```
enricobottazzi, line 49: Can we store the `vk` somewhere else? Because the `vk` for the Inclusion circuit is always the same but in this way we are storing n duplicates of the `vk` for each inclusion proof

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        for asset in &self.assets {
+            let index = asset_names.iter().position(|x| *x == asset.name).unwrap();
+            assets_sum[index] = asset.sum_balances;
+        }
+
+        // generate solvency proof
+        // TODO: fetch pk from outside. For now, we generate them here
+        let circuit = SolvencyCircuit::<L, N_ASSETS, N_BYTES>::init_empty();
+        let params = generate_setup_params(K);
```
enricobottazzi, line 142: Same as before, I think that `pk` of Solvency Circuit can be generated at `init()` and stored somewhere in the `Snapshot`

enricobottazzi (changes requested):

backend/src/apis/snapshot_data.rs
sifnoc: My bad, Fixed at [0f0a098](https://github.com/summa-dev/summa-solvency/pull/88/commits/0f0a0983908b2636763fdc2defc22f7177ad7b7f)

backend/src/apis/snapshot_data.rs
sifnoc: Yes, I know but the `root_hash` is already in the `mst`. 
I would prefer to avoid adding it redundantly.

backend/src/apis/snapshot_data.rs
sifnoc: For checking with error message on the test.
Fixed at [0f0a098](https://github.com/summa-dev/summa-solvency/pull/88/commits/0f0a0983908b2636763fdc2defc22f7177ad7b7f)

backend/src/apis/snapshot_data.rs
sifnoc: To avoid duplicated, same as `InclusionProof`

backend/src/apis/snapshot_data.rs
sifnoc: the generation of proofs using the aggregated circuit is quite time-consuming, 
so I was considering executing it in the background in the future. 

Now, we are using a solvency proof for that, I think it's okay to run when initiating this struct.
But, Let's change it later, as I am keen on gauging the duration of execution within the 'new' method through CLI at a later stage.

backend/src/apis/snapshot_data.rs
sifnoc: Ah I got it
Fixed at [0f0a098](https://github.com/summa-dev/summa-solvency/pull/88/commits/0f0a0983908b2636763fdc2defc22f7177ad7b7f)

backend/src/apis/snapshot_data.rs
sifnoc: Does the `pk` is needed for the on-chain verifier?

backend/src/apis/snapshot_data.rs
enricobottazzi: Agree

backend/src/apis/snapshot_data.rs
enricobottazzi: Yes, it is needed!

backend/src/apis/snapshot_data.rs
sifnoc: I see, I created `MstParamsAndKeys` for it
Updated at [309c563](https://github.com/summa-dev/summa-solvency/pull/88/commits/309c563d3c7862aee279b07ee035095c198c3792)

backend/src/apis/snapshot_data.rs
sifnoc: I believe it's acceptable to generate the solvency proof upon receiving a request through the 'generate_solvency_proof' method. 
In any case, I have updated the `SolvenctyProof` in [309c563](https://github.com/summa-dev/summa-solvency/pull/88/commits/309c563d3c7862aee279b07ee035095c198c3792)

backend/src/apis/snapshot_data.rs
sifnoc: I reshape all proof structs, you will satisfy it :)
Update at [309c563](https://github.com/summa-dev/summa-solvency/pull/88/commits/309c563d3c7862aee279b07ee035095c198c3792)

backend/src/apis/csv/assets_2.csv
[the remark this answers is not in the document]
sifnoc: Sure! Create issue: [Consider fetching balances for Improved Efficiency](https://github.com/summa-dev/summa-solvency/issues/92)

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,286 @@
...
+    proof_of_solvency: Option<SolvencyProof<N_ASSETS>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct Asset {
+    pub name: String,
+    pub pubkeys: Vec<String>,
+    pub balances: Vec<BigInt>,
+    pub sum_balances: Fp,
+    pub signatures: Vec<String>,
```
alxkzmn, line 50: Currently, the contract will take the exchange addresses and scan all of them to calculate the total sum of a particular asset, but the signature check is a separate process. Since the contract takes the CEX addresses as an independent argument, it simply checks each address' signature while iterating over the addresses. Do you think that it makes sense to attach signatures to an asset here? This way multiple assets could have the same set of signatures.

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,286 @@
...
+mod tests {
+    use super::*;
+
+    const N_ASSETS: usize = 2;
+    const L: usize = 2 + (N_ASSETS * 2);
+    const LEVELS: usize = 4;
+    const N_BYTES: usize = 31;
+    const K: u32 = 11;
+
+    fn initiate_snapshot_data() -> SnapshotData<LEVELS, L, N_ASSETS, N_BYTES, K> {
```
alxkzmn, line 242: "Initialize" would be a better wording here.

alxkzmn (changes requested):

backend/src/apis/snapshot_data.rs
sifnoc: I got your point! You are right, we need to separate the signatures from here to avoid storing duplicate signatures.

backend/src/apis/snapshot_data.rs
sifnoc: Fixed at [771ad22](https://github.com/summa-dev/summa-solvency/pull/88/commits/771ad223da27f5dd9d15e5e3d63d644aae4289f9)

backend/src/apis/snapshot_data.rs
sifnoc: Updated at [771ad22](https://github.com/summa-dev/summa-solvency/pull/88/commits/771ad223da27f5dd9d15e5e3d63d644aae4289f9)

alxkzmn (approved):

enricobottazzi (changes requested):

backend/src/apis/snapshot_data.rs
enricobottazzi: I still see an issue here. 

The smart contract verifier should be already deployed when the CEX calls `generate_solvency_proof`. In particular, the smart contract requires the `vk` of the Circuit in order to be deployed. Therefore, it will only accept proof generated from the corresponding `pk`. 

In the current implementation `vk` and `pk` are generated inside `generate_solvency_proof` so the proof won't get verified by the smart contract because it is generated using a wrong proving key. 

I suggest here an implementation that I hope would make our life easier:

- Replace `MstParamsAndKeys` struct but a more abstract struct `CircuitArtifacts` that stores `pk` and, optionally, `vk` of a Circuit. Since it can store any proving and verification key, it no longer needs any generic parameters such as LEVELS, L ...
- Modify `new` to fetch `pk` for `MstInclusionCircuit` from storage, create a new instance of `CircuitArtifacts` and add it to Snapshot data. The `vk` can be retrieved from `pk` and added to the instance too. Then do the same for `SolvencyCircuit`, here we don't need the `vk`
- Since we don't have to call `generate_setup_params` any more, we can remove at all the need for `K` and any issue related to that. 

Note that in the latest commit to #80, I added the `pk` for the circuits to the `backend/artifacts` folder. In particular, you can see that the file name is based on the value assigned to the generics of the circuit. For example `mst_inclusion_15_6_2.pk` is the proving key for a `MstInclusionCircuit` of LEVELS=15, L=6, N_ASSETS=2

enricobottazzi: There's still a problem here. If we generate a new random params here, the prover will fail, since the `pk` that we are fetching from `pk_path` was generated using some different params. We should make sure that the `params` used to deploy the verifier contract are the same as the ones used to extract the `pk` used to generate the proof

enricobottazzi (changes requested):

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,312 @@
...
+    proof: Vec<u8>,
+}
+
+pub struct InclusionProof {
+    // public inputs
+    leaf_hash: Fp,
+    root_hash: Fp,
+    // need for verification
+    vk: VerifyingKey<G1Affine>,
+    pk: ProvingKey<G1Affine>,
```
enricobottazzi, line 70: I don't think we need to store `pk` to InclusionProof.

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+#[derive(Debug, Clone)]
+struct SolvencyProof<const N_ASSETS: usize> {
+    // public inputs
+    root_hash: Fp,
+    assets_sum: [Fp; N_ASSETS],
+    // need for verification
+    // vk is already exist in the onchain verifier
+    pk: ProvingKey<G1Affine>,
```
enricobottazzi, line 81: Again, I don't think we need to store `pk` here, since it's never used

backend/src/apis/snapshot_data.rs
```diff
@@ -0,0 +1,312 @@
...
+        solvency::SolvencyCircuit,
+        utils::{full_prover, generate_setup_params},
+    },
+    merkle_sum_tree::utils::big_int_to_fp,
+    merkle_sum_tree::MerkleSumTree,
+};
+
+use crate::apis::csv_parser::parse_csv_to_assets;
+
+struct MstParamsAndKeys<const LEVELS: usize, const L: usize, const N_ASSETS: usize> {
```
enricobottazzi, line 25: Are the generics needed here? 
```suggestion
struct MstParamsAndKeys {
    ...
}
```

sifnoc: My bad, No needed

backend/src/apis/snapshot_data.rs
sifnoc: Updated at 71d4b19e5a87d3d236622d32bb9831840767be00

backend/src/apis/snapshot_data.rs
sifnoc: Fixed at 71d4b19e5a87d3d236622d32bb9831840767be00

backend/src/apis/snapshot_data.rs
sifnoc: Fixed at 71d4b19e5a87d3d236622d32bb9831840767be00

enricobottazzi (approved):

PR #99, opened 2023-07-12
Implement the signer

alxkzmn:


backend/Cargo.toml
```diff
@@ -5,10 +5,14 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[[example]]
```
sifnoc, line 8: "Have you considered relocating `contract_interface_gen.rs` to the scripts directory and configuring `example` to `bin`? This could be more intuitive.

backend/src/signer/signer.rs
```diff
@@ -0,0 +1,125 @@
...
+    ) -> Self {
+        let wallet: LocalWallet = LocalWallet::from_str(main_signer_key).unwrap();
+
+        let provider = Provider::<Http>::try_from(rpc_url).unwrap();
+        let client = Arc::new(SignerMiddleware::new(
+            provider,
+            wallet.with_chain_id(chain_id),
+        ));
+
+        // An example of how to retrieve the contract address from the deployments.json file
```
sifnoc, line 36: Regarding my previous comment, I believe this is a good example to place in the `example` folder. 
For now, we can just leave it as is.

backend/src/signer/signer.rs
```diff
@@ -0,0 +1,125 @@
...
+        Self {
+            signing_wallets: private_keys
+                .iter()
+                .map(|private_key| LocalWallet::from_str(private_key).unwrap())
+                .collect(),
+            summa_contract: contract,
+        }
+    }
+
+    fn read_payload_from_file<P: AsRef<Path>>(path: P) -> Result<Value, Box<dyn Error>> {
```
sifnoc, line 57: This function is only used in the example that is commented above.

backend/src/signer/signer.rs
```diff
@@ -0,0 +1,125 @@
...
+        let reader = BufReader::new(file);
+
+        // Read the JSON contents of the file
+        let u: Value = serde_json::from_reader(reader)?;
+
+        Ok(u)
+    }
+
+    async fn sign_message(wallet: &LocalWallet, exchange_id: &str) -> Signature {
+        let encoded_message = encode(&[
```
sifnoc, line 69: I believe we no longer use `exchange_id` in the message. 
Also, could you consider retrieving this message from an environment variable (i.e., .env)?"

backend/Cargo.toml
alxkzmn: I'll take a look into that, thanks!

backend/src/signer/signer.rs
alxkzmn: I can remove it, but I guess you'll need to use something similar to fetch the deployment address somewhere in snapshot to initialize the signer with that value.

backend/src/signer/signer.rs
alxkzmn: Sure, will refactor.

backend/src/signer/signer.rs
sifnoc: Ah, good point! Yes, you are right, could you make it public?

backend/src/apis/fetch.rs
```diff
@@ -115,6 +115,7 @@ mod tests {
     use super::*;
 
     #[test]
+    #[ignore]
```
enricobottazzi, line 118: If I remove the ignore and add an .env file with `COVALENT_API_KEY=cqt_rQYxVkytpYCwgC9XmPrkBq8HgVk6:`
I get this tests working. 

Can you try reproducing it on your machine? The only reason that it might be failing is that they enforce some check on the IP that uses the API, so maybe it's worth also trying using a different API KEY

backend/src/signer/signer.rs
```diff
@@ -0,0 +1,122 @@
...
+
+#[derive(Debug)]
+pub struct SummaSigner {
+    signing_wallets: Vec<LocalWallet>,
+    summa_contract: Summa<SignerMiddleware<Provider<Http>, LocalWallet>>,
+}
+
+impl SummaSigner {
+    pub fn new(
+        private_keys: &[&str],
```
enricobottazzi, line 22: Can you specify in the comments what's the difference between `private_keys` and `main_signer_key`

backend/src/signer/signer.rs
```diff
@@ -0,0 +1,122 @@
...
+        Self {
+            signing_wallets: private_keys
+                .iter()
+                .map(|private_key| LocalWallet::from_str(private_key).unwrap())
+                .collect(),
+            summa_contract: contract,
+        }
+    }
+
+    pub fn get_deployment_address<P: AsRef<Path>>(
```
enricobottazzi, line 46: What is the role of this function? I don't see it used anywhere

enricobottazzi (changes requested):
Looks neat! 

One more thing is that I get a lot of clippy warnings when building. Can you fix those?

backend/src/signer/signer.rs
alxkzmn: It will be used in backend to get the deployed Summa address from the json file.

alxkzmn:
> One more thing is that I get a lot of clippy warnings when building. Can you fix those?
Resolved a couple that were valid, but the majority of them are because the signer is not yet used by the backend code so Rust treats it al as unused. They should disappear as soon as it is integrated.

enricobottazzi:
LGTM! Just re running the workflow. Not sure why it is not fetching the Covalent API from the secrets. Hopefully we'll get rid of this very soon 😅

enricobottazzi (approved):
LGTM! Just re running the workflow. Not sure why it is not fetching the Covalent API from the secrets. Hopefully we'll get rid of this very soon 😅

sifnoc (approved):

PR #103, opened 2023-07-13
`zk-prover` added features [LOW PRIO]

enricobottazzi:
- the Poseidon Hasher generic parameters `RATE` and `WIDTH` are set to 1 and 2 in the `MerkleSumTree`, the `MstInclusionCircuit` and the `SolvencyCircuit`. 
- the range check is being performed on the root balances inside the `MstInclusionCircuit`. That's because we need to make sure that the balances are in the 64 bits range in order to perform the `lt` check. In fact, the `LtChip` needs to be parametrized by N_BYTES, which is the range in which `lhs` and `rhs` should lie. Without performing the range check on root balances, there's a risk that the value would exceed the 64 bits range and therefore creating a bug in the `lt` check
https://github.com/summa-dev/summa-solvency/blob/6d46504b74a00f75203e1fe42b63fe5ad5db4651/zk_prover/src/chips/less_than/less_than_vertical.rs#L33
- the `OverflowChip` takes a single generic `N_BYTES` similarly to how the `LtChip` work.
- the toolchain has moved to `nightly-2023-07-11` to support `#![feature(generic_const_exprs)]` which allows to use expressions inside const generics. This allowed us to remove `L` as a generic of our circuit, since it can be safely derived from `N_ASSETS`. Other generics that have been removed are `L_NODE` and L_ENTRY` in the `MerkleSumTree`


alxkzmn (approved):

zk_prover/src/circuits/merkle_sum_tree.rs
```diff
@@ -37,8 +34,10 @@ pub struct MstInclusionCircuit<const LEVELS: usize, const L: usize, const N_ASSE
     pub root_hash: Fp,
 }
 
-impl<const LEVELS: usize, const L: usize, const N_ASSETS: usize> CircuitExt<Fp>
-    for MstInclusionCircuit<LEVELS, L, N_ASSETS>
+impl<const LEVELS: usize, const N_ASSETS: usize> CircuitExt<Fp>
+    for MstInclusionCircuit<LEVELS, N_ASSETS>
+where
+    [(); 2 * (1 + N_ASSETS)]: Sized,
```
sifnoc, line 40: Is there a specific reason for using the `unit` type here? 
If it's not necessary, could you please consider changing it to `usize` instead of `()` in every `where` clause?

sifnoc (changes requested):
Thank you for implementing this feature. I believe this will significantly reduce our concerns about generics. 

I do want to mention a future consideration: We should use const generics in a clear and straightforward way within the 'where' clause, avoiding the complex usage demonstrated in the example from https://github.com/rust-lang/project-const-generics/issues/37. This is just a small point I wanted to highlight to prevent potential issues down the line.

zk_prover/src/circuits/merkle_sum_tree.rs
enricobottazzi: Good point! Fixed it in b089b09.

I also applied many fixes to `backend` when merging and resolving conflicts from master in 1eef622. Can you have a look to that?

I think we should merge #116 first and then we can merge this!

sifnoc (approved):

PR #110, opened 2023-07-19
Remove redundant Ownable from Summa

### LINKED ISSUE #109: Redundant Ownable
`Ownable.sol` is inherited in `Summa.sol`, yet it is not being used.
https://github.com/summa-dev/summa-solvency/blob/ded7155c585eabef653bfc84ec26ca10a6acd632/contracts/src/Summa.sol#L7

alikonuk1:
Like mentioned in #109 , Ownable seems redundant in `Summa.sol`
Closes #109

### DIFF
--- a/contracts/src/Summa.sol
+++ b/contracts/src/Summa.sol
@@ -4,7 +4,6 @@ pragma solidity ^0.8.18;
 // Uncomment this line to use console.log
 //import "hardhat/console.sol";
 
-import "@openzeppelin/contracts/access/Ownable.sol";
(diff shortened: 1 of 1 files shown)


alxkzmn:
It's definitely useless now but I'm afraid that if we remove it now we'll re-introduce it not later than next week when the new contract architecture will be implemented. In particular, adding or modifying the address ownership verifiers and asset balance retrievers will require admin privileges. What do you think @enricobottazzi ?

enricobottazzi:
It looks ok to be merged for now! 

Opening a new issue at #115 to what @alxkzmn mentioned

PR #114, opened 2023-07-19
Optimize the maintenance of `cexAddresses` array

alikonuk1:
As mentioned here;
https://github.com/summa-dev/summa-solvency/blob/ded7155c585eabef653bfc84ec26ca10a6acd632/contracts/src/Summa.sol#L60

The existing approach of maintaining the `cexAddresses` array can be optimized for gas efficiency. The current method removes excess elements from the `cexAddresses` array one-by-one, which could be costly in terms of gas when the array is significantly larger than `_cexAddresses`.

So an alternative approach can be to create a new array with the correct size and replace the old array with it as a whole. 
Or even just replace the old array with the input `_cexAddresses` array directly.

### DIFF
--- a/contracts/src/Summa.sol
+++ b/contracts/src/Summa.sol
@@ -41,12 +41,9 @@ contract Summa is Ownable {
             "CEX addresses and signatures count mismatch"
         );
 
-        for (uint i = 0; i < _cexAddresses.length; i++) {
-            if (i >= cexAddresses.length) {
-                cexAddresses.push(_cexAddresses[i]);
-            } else if (_cexAddresses[i] != cexAddresses[i]) {
-                cexAddresses[i] = _cexAddresses[i];
-            }
+        address[] memory newCexAddresses = new address[](_cexAddresses.length);
+
+        for (uint i = 0; i < _cexAddresses.length; ++i) {
             address recoveredPubKey = keccak256(abi.encode(message))
                 .toEthSignedMessageHash()
                 .recover(cexSignatures[i]);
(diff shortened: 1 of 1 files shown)


alxkzmn:
We've been baking something in the background that's going to be implemented next week:
```
mapping(bytes32 => bool) public confirmedAddress;
...
function submitProofOfAccountOwnership(
        OwnedAddress[] memory _cexAddresses
    ) public {
        for (uint i = 0; i < _cexAddresses.length; i++) {
            bytes32 addressHash = keccak256(_cexAddresses[i].assetAddress);
            if (!confirmedAddress[addressHash]) {
                confirmedAddress[addressHash] = true;
            }
            require(
                assetAddressVerifier[_cexAddresses[i].assetType]
                    .verifyAddressOwnership(
                        _cexAddresses[i].assetAddress,
                        _cexAddresses[i].ownershipProof
                    ),
                "Invalid signer for ETH address"
            );
        }

        emit ExchangeAddressesSubmitted /*TBD*/();
    }
```
So the current solution will be obsolete.

alxkzmn:
CC @enricobottazzi

enricobottazzi:
Thanks @alikonuk1 for the proposed implementation! I'll close the PR as it seems the design of the Smart Contract is under refactoring at the moment. 

@alxkzmn, if you have any another required features or optimization required at the smart contract level please add issues for that such that maybe Ali can help. 

Also if you'd like to get in touch with us before working on any issue you can use the `summa` channel here => https://discord.gg/EXZpaMJM

PR #116, opened 2023-07-27
Removing Covalent API and retrieving balances through provider

sifnoc:
Related issue https://github.com/summa-dev/summa-solvency/issues/108

This PR enhances the flexibility between `snapshot` and `signer` through the removal of the Covalent API

Changes:
- The `fetch_asset_sums` now requires instances, `client` and `contracts`. 
	 - The client is a `SignerMiddleware`, which must be initiated outside the function.
	 	```Rust
	 	pub async fn fetch_asset_sums<'a, M: Middleware + 'a>(
		    client: Arc<SignerMiddleware<Provider<Http>, LocalWallet>>,
		    contracts: Vec<Box<dyn BalanceFromContract<M> + Send>>,
		    asset_addresses: Vec<H160>,
		) -> Result<Vec<U256>, Box<dyn Error>> {
		```
	 - The contracts input is a vector of `Contract` from `ethers-rs` that implements the `BalanceFromContract` trait. As seen in the test cases, this facilitates support for various ERC contracts to fetch balances.
- The `generate_proof_of_solvency` function now requires `asset_sums` that are pre-fetched balances.
  ```Rust
	pub fn generate_proof_of_solvency(
      &self,
      asset_contract_addresses: Vec<String>,
      asset_sums: [Fp; N_ASSETS],
   ```


backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
+    prelude::SignerMiddleware,
+    providers::{Http, Middleware, Provider},
+    signers::LocalWallet,
+    types::{H160, U256},
+};
+
+use crate::contracts;
+use contracts::generated::mock_erc20::MockERC20;
+
+pub trait BalanceFromContract<M: Middleware> {
```
enricobottazzi, line 16: What is this trait for?

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
-        if let Some(item) = response.data.items.iter().find(|&item| {
-            item.contract_address.to_ascii_lowercase() == contract_adddress.to_ascii_lowercase()
-        }) {
-            match item.balance.parse::<u64>() {
-                Ok(num) => balances.push(num),
-                Err(e) => println!("Failed to parse string: {}", e),
-            }
-        } else {
-            balances.push(0 as u64);
+pub async fn fetch_asset_sums<'a, M: Middleware + 'a>(
```
enricobottazzi, line 21: Can you explain what are the inputs of such function? 

The way it worked before was `address` is the wallet address we are querying the `asset_sums` and `asset_contract_addresses` were the ERC20 token contract addresses

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
+    client.get_block_number().await?;
+
+    let mut result: Vec<U256> = Vec::new();
+
+    let mut get_balance_futures = Vec::new();
+    for addr in asset_addresses.clone() {
+        get_balance_futures.push(client.get_balance(addr, None));
+    }
+    let get_balances = try_join_all(get_balance_futures).await?;
+    let sum_eth_balance = get_balances
```
enricobottazzi, line 36: Can we specify in the comments that the first balance returned is `eth`?

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
         }
+        result.push(sum_erc_balance)
     }
-    Ok(balances)
+
+    Ok(result)
 }
 
 #[cfg(test)]
 mod tests {
```
enricobottazzi, line 61: I think that this test will always fail if we don't have a local chain running. Namely every github action will fail. We may add `ignore` keyword to it

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
+    use crate::contracts::tests::initialize_anvil;
 
-        let usdc_contract = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48".to_string();
-        let eth_contract = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee".to_string();
+    use ethers::utils::Anvil;
 
-        let balances =
-            fetch_balances_per_addr(address, [usdc_contract, eth_contract].to_vec()).unwrap();
+    #[tokio::test]
+    async fn test_fetch_asset_sums() {
```
enricobottazzi, line 69: I don't understand this test. Are the 2 `asset_sums` returned related to 2 different erc20 contracts. Or is it one `eth` and the other `mockerc20` ?

If so how are you able to set the balance of the `cex_addr_1` and `cex_addr_2` inside the erc20?

backend/README.md
```diff
@@ -25,7 +25,7 @@ wget https://trusted-setup-halo2kzg.s3.eu-central-1.amazonaws.com/hermez-raw-11
 
 and pass the path to the file to the `Snapshot::new` method.
 
-Furthermore, the `generate_proof_of_solvency` method requires to fetch data about the balances of the wallets of the CEX. This data is fetched using the Covalent API. In order to use this method, you need to create an `.env` file and store the `COVALENT_API_KEY` there. You can get an API key at https://www.covalenthq.com/platform/.
+Furthermore, the `generate_proof_of_solvency` method requires to fetch data about the balances of the wallets of the CEX. the default node url is `http://localhost:8545`.
```
enricobottazzi, line 28: This is incorrect. As of now the `generate_proof_of_solvency` method doesn't fetch data at all. We can add here in the readme a description on how to use `fetch_asset_sums` to fetch data from the blockchain. Also we can provide guidance on how to test it (if we set the test to `ignore` by default)

backend/src/contracts/tests.rs
```diff
@@ -1,24 +1,69 @@
...
+use ethers::{
+    prelude::{ContractFactory, SignerMiddleware},
+    providers::{Http, Provider},
+    signers::{LocalWallet, Signer},
+    types::{H160, U256},
+};
+
+use crate::contracts::generated::mock_erc20::{MockERC20, MOCKERC20_ABI, MOCKERC20_BYTECODE};
+
+pub async fn initialize_anvil(
```
enricobottazzi, line 12: Can you specify in the comments what's happening here? What are you initializing at?

backend/src/apis/snapshot.rs
```diff
@@ -280,8 +262,13 @@ mod tests {
             "0x220b71671b649c03714da9c621285943f3cbcdc6".to_string(), // ERC20 token address
         ];
 
-        let calldata = snapshot
-            .generate_proof_of_solvency(asset_addresses.clone(), None)
+        // In this test, we assume that the balances of the accounts in the snapshot are 556863 for both assets
```
enricobottazzi, line 265: Can we remove the `ignore` for this test?

enricobottazzi (changes requested):
Hi Jin! I left a few comments, mostly doubts that I have. I think that the review of @alex will matter more than mine as these is more related to the smart contract side.

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
+    let get_balances = try_join_all(get_balance_futures).await?;
+    let sum_eth_balance = get_balances
+        .into_iter()
+        .reduce(|acc, balance| acc + balance)
+        .unwrap();
+    result.push(sum_eth_balance);
+
+    // Most in case, the number of contracts is less than the number of asset addresses
+    // Iterating over contracts first is more efficient
+    let mut sum_erc_balance = U256::zero();
```
alxkzmn, line 44: "erc" here can be misleading for someone unfamiliar with the logic. Right now it is specifically erc20 but if you want to avoid being specific you can call it "token_balance" to distinguish with, e.g., "ether_balance".

backend/src/apis/fetch.rs
alxkzmn: I believe this one is for retrieving the non-ETH balances. In this case, though, I'd prefer it to be called TokenBalance with a function "get_token_balance" because we're dealing with assets but not some arbitrary smart contracts.

backend/src/apis/fetch.rs
alxkzmn: Do you mean the old test? The new test is launching its own Anvil so there shouldn't be problems.

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
-        let usdc_contract = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48".to_string();
-        let eth_contract = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee".to_string();
+    use ethers::utils::Anvil;
 
-        let balances =
-            fetch_balances_per_addr(address, [usdc_contract, eth_contract].to_vec()).unwrap();
+    #[tokio::test]
+    async fn test_fetch_asset_sums() {
+        // Necessary to implement `get_balance_from_contract` for the `contracts` parameter by following trait
+        impl<M: Middleware> BalanceFromContract<M> for MockERC20<M> {
```
alxkzmn, line 71: I think it's safe to promote this trait implementation to live outside the test since it's going to stay the same for the case of ERC20 :)

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
-    fn test_fetch_asset_sums() {
-        let address_1 = "0xe4D9621321e77B499392801d08Ed68Ec5175f204".to_string(); // this is an address with 0 usdc and 0.010910762665574143 ETH
-        let address_2 = "0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1".to_string(); // this is an address with 0.000001 USDC usdc and 1 wei
+        let (cex_addr_1, cex_addr_2, client, mock_erc20) = initialize_anvil(&anvil).await;
 
-        let usdc_contract = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48".to_string();
-        let eth_contract = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee".to_string();
+        // send RPC requests with `anvil_setBalance` method via provider to adjust balance of `cex_addr_1` and `cex_addr_2`
+        for addr in [cex_addr_1, cex_addr_2].to_vec() {
+            let _res = client
```
alxkzmn, line 85: I think setting the ETH balances can also go into the `initialize_anvil` function, since you're already initializing the ERC20 balances there.

backend/src/contracts/tests.rs
```diff
@@ -1,24 +1,69 @@
...
-        }, signer::SummaSigner},
+        },
+        signer::SummaSigner,
+        tests::initialize_anvil,
     };
 
-
     #[tokio::test]
     async fn test_sign_message() {
         let anvil = Anvil::new().spawn();
```
alxkzmn, line 69: I think this can also go into the initialization function since it's used in multiple places and the mnemonic is the same.

contracts/scripts/deploy.ts
```diff
@@ -20,7 +20,21 @@ async function main() {
...
+  for (let i = 1; i < 4; i++) {
+    console.log(signers[i].address);
+    await ethers.provider.send("hardhat_setBalance", [signers[i].address, "0x2d515"])
+  }
+
+  const mockERC20 = await ethers.deployContract("MockERC20");
+  await mockERC20.deployed();
+
+  console.log(`MockERC20 deployed to ${mockERC20.address}`);
+  await mockERC20.mint("0x90F79bf6EB2c4f870365E785982E1f101E93b906", 556863);
```
alxkzmn, line 35: Is it possible to keep all the mock balance assignment and minting within the tests? Something like a single `initialize_anvil` function would suffice. Usually, a separate deployment script is written as a general-purpose one, assuming it can be used for real deployments, so it wouldn't do the mock minting.

alxkzmn (changes requested):

backend/src/apis/fetch.rs
sifnoc: > What is this trait for?

It's a trait designed to wrap the `ERC` token contract interfaces (e.g., `MockERC20`) that are generated by `abigen`.

For instance, consider two token contracts, `MockERC20_A` and `MockERC20_B`. Each contract interface file has a balance_of method. However, the `fetch_asset_sums` function could not directly call the `balance_of` function without utilizing an enum pattern. But with this pattern, we would have to add both `MockERC20_A` and `MockERC20_B` to the enum. If the CEX supports additional tokens, the enum type would become increasingly lengthy.

Instead of using enum, I implemented the `BalanceFromContract` trait for the `contracts` inputs. While this approach still requires some code modifications, we can potentially reduce the task by using `abigen` to generate the necessary bindings (interface files).

> I believe this one is for retrieving the non-ETH balances. In this case, though, I'd prefer it to be called TokenBalance with a function "get_token_balance" because we're dealing with assets but not some arbitrary smart contracts.

 I appreciate your suggestions and thank you for helping with the naming!

backend/src/apis/fetch.rs
sifnoc: > I think that this test will always fail if we don't have a local chain running. Namely every github action will fail. We may add `ignore` keyword to it

As noted by Alex, there is no problem when testing on the Anvil test chain in a local environment instead of using the headhat node. 
However, an error occurs in the GitHub action due to the mock ERC20 token balance. I will address and fix this issue.

backend/src/apis/fetch.rs
sifnoc: Sure! Extend previous comment https://github.com/summa-dev/summa-solvency/pull/116#discussion_r1279164154

we now have three inputs in the `fetch_asset_sum` function: `client`, `contracts`, and `asset_addresses`.

You can think of the `client` as a provider responsible for fetching ETH balances. Meanwhile, `contracts` is an array of `ContractInstance`, with each element representing a high-level interface connected to a specific contract address via the provider. This interface allows for easy interaction with the smart contract.

Both `client` and `contracts` fetch balances by iterating through `asset_addresses`.

I chose to use `contracts` as an input because it offers more flexibility to adapt to different types of ERC contracts, not just ERC20.

backend/src/apis/fetch.rs
sifnoc: > I don't understand this test. Are the 2 `asset_sums` returned related to 2 different erc20 contracts. Or is it one `eth` and the other `mockerc20` ?

It returns the balance of both ETH and a mock ERC20 token.

> If so how are you able to set the balance of the `cex_addr_1` and `cex_addr_2` inside the erc20?

the assets are configured within the [`initialize_anvil`](https://github.com/summa-dev/summa-solvency/blob/abf920fccc4818a43fce00a142b4a5ba6c12c82a/backend/src/contracts/tests.rs#L12-L46) function

https://github.com/summa-dev/summa-solvency/blob/abf920fccc4818a43fce00a142b4a5ba6c12c82a/backend/src/contracts/tests.rs#L12-L46

For the purpose of this test, you can think of the test chain as `Anvil` instead of `Hardhat`

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
-            }
-        } else {
-            balances.push(0 as u64);
+pub async fn fetch_asset_sums<'a, M: Middleware + 'a>(
+    client: Arc<SignerMiddleware<Provider<Http>, LocalWallet>>,
+    contracts: Vec<Box<dyn BalanceFromContract<M> + Send>>,
+    asset_addresses: Vec<H160>,
+) -> Result<Vec<U256>, Box<dyn Error>> {
+    // For checking connectivity
+    client.get_block_number().await?;
```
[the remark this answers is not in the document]
sifnoc: Yeah, It makes sense. I will remove it from the function.

backend/src/contracts/tests.rs
sifnoc: I will update it while fixing error on github action

backend/README.md
sifnoc: Good point!

I will update the README to explain the different methods available for fetching balances. 

There are two primary ways to accomplish this:

You can fetch balances using the `fetch_asset_sums` function.
Alternatively, you can create your own custom fetcher to retrieve the balances.

Once the balances have been obtained, you can proceed to use them in the `generate_proof_of_solvency` function.

backend/src/apis/fetch.rs
enricobottazzi: Ok thanks for clarifying it! I think it makes sense to keep the pattern as you suggested following the naming proposed by Alex!

backend/src/apis/fetch.rs
enricobottazzi: Understood!

backend/src/apis/fetch.rs
```diff
@@ -1,151 +1,102 @@
...
+
+    let mut result: Vec<U256> = Vec::new();
+
+    let mut get_balance_futures = Vec::new();
+    for addr in asset_addresses.clone() {
```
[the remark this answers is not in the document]
enricobottazzi: Yes I agree with this, I think that also `contracts` might be well replaced by `token_contracts` if we don't want to limit it to `erc_20_token_contracts`

contracts/scripts/deploy.ts
sifnoc: Ah I see, we can remove this and the cli app also using `Anvil` instead of `hardhat`.
Thanks for point out

backend/src/apis/fetch.rs
sifnoc: Changed names at 9ada33502519

backend/src/apis/fetch.rs
sifnoc: Updated at 9ada33502519

contracts/scripts/deploy.ts
sifnoc: It removed at 9ada335

backend/src/apis/fetch.rs
sifnoc: Moved to `initialize_anvil` at https://github.com/summa-dev/summa-solvency/commit/9ada335025190df0da7e21ebfc361ee2ab68fdd5

backend/src/apis/fetch.rs
sifnoc: Updated at 9ada335

backend/src/apis/fetch.rs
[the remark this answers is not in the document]
sifnoc: Updated at 9ada335

backend/src/apis/fetch.rs
sifnoc: I create an issue for this.
https://github.com/summa-dev/summa-solvency/issues/123

backend/src/apis/snapshot.rs
sifnoc: Removed at 9ada335

backend/README.md
sifnoc: Updated at [eec59ee](https://github.com/summa-dev/summa-solvency/pull/116/commits/eec59ee96add41075e44f6c9b1c7efa94ffdb361)

backend/src/contracts/tests.rs
sifnoc: Updated at 8502144c

backend/src/apis/fetch.rs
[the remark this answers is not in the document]
sifnoc: Removed at 8502144c

backend/src/contracts/tests.rs
sifnoc: Create `AnvilInstance` in the `initialize_anvil` method 8502144c

backend/README.md
```diff
@@ -13,19 +13,61 @@ Furthermore, the `Snapshot` struct contains the following methods:
...
+
+```Rust
+let asset_sums = fetch_asset_sums(client, token_contracts, exchange_addresses).await?;
+```
+
+Please note that the first element in the `asset_sums` array represents the ETH balance.
+
+Alternatively, you can create your own custom fetcher to retrieve the balances.
+
+### Proof of Ownership
```
enricobottazzi, line 44: In this section we explain how to generate signatures in order to generate a `proof_of_account_ownership`. But actually the signatures needs to be provided before, when initializing the `Snapshot` within a `csv` file.

So maybe it's better to remove the `proof of ownership` section and add a section to the readme that explains what are the input needed to initalize the `Snapshot`. When mentioning the `csv` file containing the signatures, you can mention that we have a method that helps generating a signature starting from a message and a private key. In any case the signatures are to be added to a `csv` file  to be processed into  a proof of account ownership

enricobottazzi (commented):
LGTM! Left a few minor comments related to the readme and some in-line comments, when fixed it is ready to go

backend/src/contracts/signer.rs
```diff
@@ -34,7 +36,10 @@ impl SummaSigner {
     ) -> Self {
         let wallet: LocalWallet = LocalWallet::from_str(main_signer_key).unwrap();
 
-        let provider = Provider::<Http>::try_from(rpc_url).unwrap();
+        // let provider = Provider::<Http>::try_from(rpc_url).unwrap();
```
alxkzmn, line 39: Looks like the comment is unnecessary

alxkzmn (approved):

backend/src/contracts/signer.rs
sifnoc: Removed at [70ad9c9](https://github.com/summa-dev/summa-solvency/pull/116/commits/70ad9c97f31c3573941a23cb0ff7a36952f95392)

backend/README.md
sifnoc: Removed at [70ad9c9](https://github.com/summa-dev/summa-solvency/pull/116/commits/70ad9c97f31c3573941a23cb0ff7a36952f95392)

backend/README.md
```diff
@@ -13,19 +13,61 @@ Furthermore, the `Snapshot` struct contains the following methods:
 - `generate_solvency_verifier` -> write the Solidity Verifier contract (for the `SolvencyProof`) to a file
 - `generate_proof_of_solvency` -> generate the `SolvencyProof` for the current snapshot to be verified on-chain
 - `generate_inclusion_proof` -> generate the `MstInclusionProof` for a specific user for the current snapshot to be verified off-chain
```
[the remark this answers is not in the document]
sifnoc: Fixed at [70ad9c9](https://github.com/summa-dev/summa-solvency/pull/116/commits/70ad9c97f31c3573941a23cb0ff7a36952f95392)

backend/src/apis/snapshot.rs
```diff
@@ -157,27 +157,9 @@ impl<const LEVELS: usize, const L: usize, const N_ASSETS: usize, const N_BYTES:
     pub fn generate_proof_of_solvency(
         &self,
         asset_contract_addresses: Vec<String>,
-        asset_sum: Option<[Fp; N_ASSETS]>,
+        asset_sums: [Fp; N_ASSETS],
     ) -> Result<(SolvencyProof, Vec<String>), &'static str> {
         // For each asset, identified by its contract address, fetch the sum of all balances owned by the accounts contained in the snapshot
```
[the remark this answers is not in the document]
sifnoc: Removed at [70ad9c9](https://github.com/summa-dev/summa-solvency/pull/116/commits/70ad9c97f31c3573941a23cb0ff7a36952f95392)

enricobottazzi (approved):

PR #124, opened 2023-08-01
fix: `swap_constraint` bug

enricobottazzi:
This PR fixes a bug in our current `swap_constraint`.

In particular the current swap_constraint `s * swap_bit * ((elelment_l_next - elelment_l_cur) - (elelment_r_cur - elelment_r_next))` correctly check that the swapping is performed correctly **only** in the case in which `swap_bit` is equal to 1 and therefore the swapping needs to be performed. 

In the case in which `swap_bit` is equal to 0, meaning that the swapping between row(cur) and row(next) doesn't need to be performed, the polynomial constraint will always be equal to 0, therefore without checking that the values were carried from row(cur) to row(next) correctly without performing any swap.

The new polynomial constraint solves this issue

### DIFF
--- a/zk_prover/src/chips/merkle_sum_tree.rs
+++ b/zk_prover/src/chips/merkle_sum_tree.rs
@@ -20,7 +20,8 @@ pub struct MerkleSumTreeConfig {
 /// Chip that performs various constraints related to a Merkle Sum Tree data structure such as:
 ///
 /// * `s * swap_bit * (1 - swap_bit) = 0` (if `bool_and_swap_selector` is toggled). It basically enforces that swap_bit is either a 0 or 1.
-/// * `s * swap_bit * ((elelment_l_next - elelment_l_cur) - (elelment_r_cur - elelment_r_next))`. Enforces that if the swap_bit is equal to 1, the values will be swapped on the next row (if `bool_and_swap_selector` is toggled).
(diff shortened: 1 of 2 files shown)


sifnoc (commented):
This reminds me of your comments in https://github.com/summa-dev/summa-solvency/pull/8#discussion_r1201922913 when you refactored the merkle_sum_tree.

It appears that the constraint, without doubling, is also satisfied in this narrowed version of the chip.

> therefore without checking that the values were carried from row(cur) to row(next) correctly without performing any swap.

I'm not entirely clear on the meaning of "performing" in this context. Does the code already perform a swap during the assignment phase here: 
https://github.com/summa-dev/summa-solvency/blob/7e39b84b9210ff149a545cfe54495297283edb45/zk_prover/src/chips/merkle_sum_tree.rs#L203-L209

It seems to me that both the previous and fixed constraints work well without any bugs.

sifnoc:
> | a   | b   | swap_bit   |
> | --- | --- | -    |
> |  5  |  6  | 0    |
> |  6  |  5  |  -   |  
> ```
> 
> The constraint will still be satisfied in this case, which shouldn't be the case! `1 * 0 * ((6-5) - (6-5)) = 0` white_check_mark
> 
> So, to conclude, the bug in the current version is that it is not constraining the swap to not be performed when `swap_bit` is set to 0

Thanks for the kind explanation! Oh, I see now, it's a bug!
The constraint can be satisfied even if the assigning logic is manipulated.

sifnoc (approved):

alxkzmn (approved):
Let's fix the "elelment" in the next PR

Issue #119, opened 2023-07-31
Fix bug on `MSTInclusionCircuit`


enricobottazzi:
As of now there's no binding between the `leaf_balances` and the `leaf_hash` making the circuit prone to exploit from malicious prover

PR #129, opened 2023-08-03
Fix entry bug

enricobottazzi:
The PR solves #119.

In particular in the current version of the `MstInclusionCircuit` there's no binding between the `leaf_hash` (public input` and the `leaf_balances[]` private input.  

Screenshot 2023-08-03 at 10 37 04

`leaf_hash` should be equal to `H(username, leaf_balances[])` but this check is not enforced inside the circuit, so a malicious prover can pass the valid `leaf_hash` of the user with an understated balance and the user won't notice it.

The fix for this is to pass the `username` as private input and compute the `leaf_hash` inside the circuit 

Screenshot 2023-08-03 at 10 37 18

The changes applied in this PR affect:

- `MerkleSumTreeChip`, in which `assign_entry_hash_and_balances` and `assing_swap_bit` functions have been deprecated in favor of a more generic `assign_value`
- `MstInclusionCircuit` in which the synthesize function follows a different logic, as described before
- `SolvencyCircuit` in which the synthesize function follows the same logic but uses a different assignment function (given the functions deprecated in `MerkleSumTreeChip` as described before.

Before opening a PR I ask you (@alxkzmn and @sifnoc) to suggest how this can affect `smart_contracts` and `backend`. The first thing that comes to my mind is that the solvency verifier smart contract needs to be redeployed according to the modified structure of the circuit.


sifnoc:
> Before opening a PR I ask you (@alxkzmn and @sifnoc) to suggest how this can affect `smart_contracts` and `backend`. The first thing that comes to my mind is that the solvency verifier smart contract needs to be redeployed according to the modified structure of the circuit.

I've reviewed it, and it seems there are no effects on the `backend`. The user still needs to perform the verification step for the `leaf_hash`, or they can generate the `leaf_hash` by themselves for public input.

enricobottazzi:
Ok, @alxkzmn should I go ahead, generate a new `SolvencyVerifier.sol` and add it to `contracts/src` ?

alxkzmn:
@enricobottazzi you can, but you will have to make sure that the tests pass in the contracts subproject. I think it should be enough to replace the proof value for that.

alxkzmn:
https://github.com/summa-dev/summa-solvency/blob/ade94981090f7359675224180de11bd40aff86ad/contracts/test/Summa.ts#L275 fortunately, seems like it's only used in one place.

zk_prover/prints/ecdsa-layout.png
alxkzmn, line 1: This is a leftover, right?

alxkzmn (changes requested):

zk_prover/prints/ecdsa-layout.png
enricobottazzi: Yes, fixed in 9aa9bdb

sifnoc (approved):
Looks good to me!

alxkzmn (approved):

PR #131, opened 2023-08-06
New smart contract architecture

alxkzmn:


contracts/src/Summa.sol
```diff
@@ -5,116 +5,159 @@ pragma solidity ^0.8.18;
...
+        bytes cexAddress;
+        bytes ownershipProof;
+    }
 
-    event ExchangeAddressesSubmitted(address[] cexAddresses);
+    /**
+     * @dev Struct representing an asset owned by the CEX
+     * @param assetType The type of the asset (e.g., a hash of the asset kind, like keccak256("ETH"), keccak256("BTC", keccak256("ERC20")
+     * @param amountToProve The amount of the asset that the CEX wants to prove to own
+     * @param addresses The addresses that the CEX wants to prove to own the asset
```
enricobottazzi, line 42:
```suggestion
     * @param addresses The addresses of which the CEX has proven ownership of that control the `asset`
```

contracts/src/Summa.sol
```diff
@@ -5,116 +5,159 @@ pragma solidity ^0.8.18;
...
-            totalETHBalance >= assetSums[0],
-            "Actual ETH balance is less than the proven balance"
-        );
-
-        for (uint i = 0; i < erc20ContractAddresses.length; i++) {
             require(
-                erc20Balances[i] >= assetSums[i + 1],
-                "Actual ERC20 balance is less than the proven balance"
+                assets[i].amountToProve <= totalAssetSum[i],
+                "Actual balance is less than the proven balance"
```
enricobottazzi, line 152:
```suggestion
                "Resulting queried balance is less than the desired amount to prove"
```

contracts/src/Summa.sol
```diff
@@ -5,116 +5,159 @@ pragma solidity ^0.8.18;
...
     }
 
     /**
      * @dev Submit proof of solvency for a CEX
-     * @param erc20ContractAddresses The addresses of the ERC20 token contracts that the CEX holds (e.g., USDT, USDC, DAI)
-     * @param assetSums The total asset sums to prove. ETH balance should be the first element, followed by ERC20 balances in the order of erc20ContractAddresses
+     * @param assets Assets owned by CEX
      * @param mstRoot The root of the Merkle sum tree
      * @param proof The ZK proof
+     * @param timestamp The timestamp at which the proof was generated
```
enricobottazzi, line 122: More precisely, `timestamp` refers to the moment in time in which the balances of the users are snapshotted in order to generate the merkle sum tree. We can say that this is the timestap at which the liabilities of the CEX are snapshotted.

contracts/test/Summa.ts
```diff
@@ -63,211 +90,184 @@ describe("Summa Contract", () => {
...
         "0x9a9f2dd5ad8242b8feb5ad19e6f5cc87693bc2335ed849c8f9fa908e49c047d0250d001da1d1a83fed254171f1c686e83482b9b927702768efdaafac7375eac91d";
 
       await expect(
-        summa.submitProofOfAccountOwnership(
-          cexAddresses,
-          cexSignatures,
-          "Summa proof of solvency for CryptoExchange"
-        )
+        summa.submitProofOfAccountOwnership(ownedAddresses)
       ).to.be.revertedWith("ECDSA: invalid signature");
```
enricobottazzi, line 180: What's the difference between invalid signature and invalid signer?

contracts/test/Summa.ts
```diff
@@ -276,144 +276,162 @@ describe("Summa Contract", () => {
...
-        "Actual ETH balance is less than the proven balance"
-      );
+      )
+        .to.emit(summa, "ProofOfSolvencySubmitted")
+        .withArgs(mstRoot);
     });
 
-    it("should revert if actual ERC20 balance is less than the proven balance", async () => {
-      balancesToProve[1] = BigNumber.from(556864);
+    it("should not verify the proof of solvency if the address verifier was not set", async () => {
```
enricobottazzi, line 344:
```suggestion
    it("should not verify the proof of solvency if the CEX hasn't proven ownership of an address", async () => {
````

File: contracts/scripts/deploy.ts [hunk 6]
```diff
@@ -51,10 +51,13 @@ async function main() {
     deploymentsStringified
   );
 
-  //Copy the ABI from `artifacts/src/Summa.sol/Summa.json` to `backend/src/contracts/Summa.json`
-  const abi = require("../artifacts/src/Summa.sol/Summa.json");
-  const abiStringified = JSON.stringify(abi);
-  fs.writeFileSync("../backend/src/contracts/Summa.json", abiStringified);
+  //Copy the ABIs from `artifacts/src/*` to `backend/src/contracts/*`
```
enricobottazzi, line 54: Can you add a reference to what you are performing with this script inside the `readme.md`

backend/build.rs
```diff
@@ -1,22 +1,48 @@
-use std::{fs::OpenOptions, io::Write, path::PathBuf};
-
 use ethers::prelude::Abigen;
+use std::{fs::OpenOptions, io::Write, path::PathBuf};
 
 fn main() {
```
enricobottazzi, line 4: Is everything mentioned in the read me according to this build script still valid after your modficiations?

enricobottazzi (commented):
The level of abstraction introduced by the `BalanceRetriever` and the `AddressVerifier` is a great improvement in our architecture! 

I left some minor comments mostly on `contracts`. For `backend` @Jin knows better.

I also made an illustration to make it clear to myself while reviewing the code => https://link.excalidraw.com/l/9X5TKqd1zLR/3CbsJR0W8Ug

contracts/src/Summa.sol
```diff
@@ -5,116 +5,159 @@ pragma solidity ^0.8.18;
...
+import "./interfaces/IAddressOwnershipVerifier.sol";
 
-contract Summa {
+contract Summa is Ownable {
     using ECDSA for bytes32;
 
     IVerifier private immutable verifier;
 
-    address[] public cexAddresses;
+    mapping(bytes32 => bool) public confirmedAddress; // instead ofbool, should contain the additional data needed to verify the address ownership (e.g., in case of ERC20, the address of the ERC20 contract)
```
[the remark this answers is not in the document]
alxkzmn: Leftover, will remove.

contracts/src/Summa.sol
```diff
@@ -5,116 +5,159 @@ pragma solidity ^0.8.18;
...
-                cexAddresses.push(_cexAddresses[i]);
-            } else if (_cexAddresses[i] != cexAddresses[i]) {
-                cexAddresses[i] = _cexAddresses[i];
-            }
-            address recoveredPubKey = keccak256(abi.encode(message))
-                .toEthSignedMessageHash()
-                .recover(cexSignatures[i]);
+            bytes32 addressHash = keccak256(_cexAddresses[i].cexAddress);
+            require(!confirmedAddress[addressHash], "Address already verified");
+            confirmedAddress[addressHash] = true;
```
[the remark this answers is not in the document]
alxkzmn: No, the transaction will revert entirely. Also, it's a good practice to modify state before the external call to avoid any hacks based on malicious external contracts, e.g., reentrancy (although we probably shouldn't worry because `verifyAddressOwnership` shouldn't be ever controlled by a malicious actor).

contracts/src/Summa.sol
```diff
@@ -5,116 +5,159 @@ pragma solidity ^0.8.18;
...
+    mapping(bytes32 => bytes) public balanceRetrieverArgs;
+    mapping(bytes32 => IAddressOwnershipVerifier) public cexAddressVerifier;
+
+    /**
+     * @dev Struct representing an address owned by the CEX
+     * @param addressType The type of the address by chain (e.g., EVM, BTC)
+     * @param cexAddress The address owned by the CEX
+     * @param ownershipProof The data needed to verify the address ownership (e.g., in case of EVM, a signature and a message signed by the address, encoded as bytes)
+     */
+    struct OwnedAddress {
```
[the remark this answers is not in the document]
alxkzmn: We can stick with addresses I guess.

contracts/test/Summa.ts
alxkzmn: Invalid signature is cryptographically invalid (impossible to recover). Invalid signer is when the signer address is not the one we expect.

sifnoc (changes requested):
Actually, there's not much effect on the backend API. Alex has already updated everything – many thanks.
However, I have requested a feature to fetch balances for multiple addresses through the Retriever contracts.

contracts/scripts/deploy.ts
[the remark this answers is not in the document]
sifnoc: > However, when we redeploy the contracts, the new ABIs get copied, and the buildscript will run when backend tests are run.

This is also good for me :)

contracts/src/Summa.sol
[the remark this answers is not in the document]
sifnoc: Ah I understood

sifnoc (approved):

Issue #130, opened 2023-08-03
More efficient `Range Check` implementation 


enricobottazzi:
The goal of the range check chip is to evaluate that a value is within the required range determined by `N_BYTES`

As mentioned in #77, the goal is to improve the performance of the range check chip such that it accomplishes the same goal by performing fewer relative rotations.

The proposed implementation can be found at => https://github.com/enricobottazzi/scaffold-halo2

The logic is the following:

This helper checks that the value witnessed in a given cell is within a given range defined by N_BYTES. We consider the example in which we want to check that `value` `0x1f2f3f4f` is in a range `N_BYTES=4` 

 |     | value      | decomposed_value | running_sum | toggle_running_sum_check | toggle_lookup_check |
 | --- | ---------- | ---------------- | ----------- | ------------------------ | ------------------- |
 | 0   | -          | -                | 0x00        | 0                        | 0                   |
 | 1   | 0x1f2f3f4f | 0x1f             | 0x1f        | 1                        | 1                   |
 | 2   |            | 0x2f             | 0x1f2f      | 1                        | 1                   |
 | 3   |            | 0x3f             | 0x1f2f3f    | 1                        | 1                   |
 | 4   |            | 0x4f             | 0x1f2f3f4f  | 1                        | 1                   |

The column decomposed_value contains the decomposed value of `value` in #N_BYTES byte chunks, big-endian.
The column running_sum contains the running sum of the values in decomposed_value. In particular the running_sum in a particular row is the result of concatenating the prev running_sum with the current decomposed_value.
The contraints that are enforced are:
 - `(running_sum(prev) << 8) + decomposed_value(cur) = running_sum(cur)` (enable by toggle_running_sum_check)
 - `decomposed_value(cur) ∈ u8_lookup_table` (enable by toggle_lookup_check)
 - `value(1) == running_sum(4)` (copy constraint applied here)

With the proposed implementation, the maximum relative rotation performed when setting the constraints equals -1.

Let's consider the case of a value `0x1f2f3f4f5f6f7f8f` that overflow the required range of `N_BYTES=4`.  This would be the behaviour:

|     | value               | decomposed_value    | running_sum              | toggle_running_sum_check | toggle_lookup_check |
|-----|-------------        |------               |------                    | ------                   | ------              |
|  0  |  -                  | -                   | 0x00                     | 0                        | 0                   |
[...]

PR #132, opened 2023-08-08
Update range check

enricobottazzi:
The PR: 

- Modifies the logic of the `RangeCheckChip` as extensively discussed in #130 
- Set `N_BYTES` as generic parameter for `MerkleSumTree`, `MstInclusionCircuit` and `SolvencyCircuit`. This parameter indicates the range in which all the balances should lie. Using it as generic would allow us to easily tune this parameter according to the requirements of a CEX. 

As a result of that we remove a series of polynomial constraints from the `RangeCheckChip` that were taking a lot of relative rotation + remove the need of an extra selector. Running some benchmarks locally, the performance in terms of proving time is still in the order of 400 ms. But we have improvement on the side of proof size (1760 -> 1568). I believe that the impact will be even bigger when working with higher `N_BYTES` since the degree of the previous polynomial constraint was a function of that value


zk_prover/src/chips/range/range_check.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Configures the Range Chip
+    pub fn configure(
+        meta: &mut ConstraintSystem<Fp>,
+        z: Column<Advice>,
+        range: Column<Fixed>,
+        toggle_lookup_check: Selector,
+    ) -> RangeCheckConfig<N_BYTES> {
+        meta.annotate_lookup_any_column(range, || "LOOKUP_MAXBITS_RANGE");
+
+        meta.lookup_any("range u8 check for word", |meta| {
```
sifnoc, line 72: In Ethereum, the term `Word` usually represents 256 bits, but the `z_0` value is less than 256 bits.
So, how about using simply `z_value` instead of `word` in this context?

zk_prover/src/chips/range/tests.rs
```diff
@@ -0,0 +1,287 @@
...
+        config: Self::Config,
+        mut layouter: impl Layouter<Fp>,
+    ) -> Result<(), Error> {
+        // Initiate the add chip
+        let addchip = AddChip::construct(config.addchip_config);
+        let (a_cell, b_cell, c_cell) =
+            addchip.assign(self.a, self.b, layouter.namespace(|| "add chip"))?;
+
+        // Initiate the range check chip
+        let overflow_chip = RangeCheckChip::construct(config.range_check_config);
```
sifnoc, line 153: How about just use `range_chip` instead of `overflow_chip`?

sifnoc (changes requested):
Great! I've left comments about a few wordings

alxkzmn (approved):

zk_prover/src/chips/range/range_check.rs
enricobottazzi: Thanks for the correction ! Fixed in 969ac61

zk_prover/src/chips/range/tests.rs
enricobottazzi: Yeah, that was a leftover from the previous version. Fixed in 969ac61

sifnoc (approved):

PR #138, opened 2023-08-18
Improvement on `LtChip` 

### LINKED ISSUE #91: `lt_gate` with fewer relative rotations
https://github.com/summa-dev/summa-solvency/blob/482f2790191be9a72ecae21cfb5b8a85757e1834/zk_prover/src/chips/less_than/less_than_vertical.rs#L82-L84

as you can see a lot of rotations are performed here. A solution would be to work with a running sum, similar to what has been proposed in #77 to make the rotation only to `next()`

enricobottazzi:
Fix #91 

A few points on this: 

- We no longer use a `LtChip` that returns 1 if `lhs<rhs` and 0 if `lhs>rhs` since we do not need to handle the second case. We now use `CheckLtChip` that constraints `lhs` to be less than `rhs` and returns an error if not. This designed is inspired by [axiom halo2 lib](https://axiom-crypto.github.io/halo2-lib/src/halo2_base/gates/range.rs.html#562)
- 
- As described in the issue, we wanted to reduce the number of relative rotations performed inside the chip. This is actually achieved very simply by performing the range check on `diff` using the [updated `RangeCheckChip`](https://github.com/summa-dev/summa-solvency/issues/130).
- According to the design of the `CheckLtChip`  a range check on `lhs` and `rhs` must be performed to avoid the check to fail.  My choice is to not perform the range check inside the `SolvencyCircuit` (which is where the `CheckLtChip` is used) since the range check on such values is already performed within the `MstInclusionCircuit`. In such scenario, a malicious exchange will be able to generate a valid π of Solvency, but it will fail any π of Inclusion. Wondering what's your opinion on this.
https://github.com/summa-dev/summa-solvency/blob/6b1ff8977bc9ca79240964519e527eb14e7def05/zk_prover/src/chips/less_than_check/lt_check.rs#L30-L32
- This improvement allowed to reduce the size of the proof and achieve support of a greater `N_BYTES` range. I moved it to 14 since according to my research the token with the biggest supply is [SHIBA INU](https://etherscan.io/token/0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce) which is 110 bits


alxkzmn (approved):

zk_prover/src/chips/less_than_check/lt_check.rs
```diff
@@ -0,0 +1,135 @@
...
+    check_lt_enable: Selector,
+    /// Configuration for the RangeCheck chip.
+    range_check_config: RangeCheckConfig<N_BYTES>,
+}
+
+/// Constrains that 'lhs' is less than 'rhs'.
+///
+/// Assumes that`lhs` and `rhs` are known to have <= N_BYTES bytes.
+///
+/// Note: This may fail silently if a or b have more than N_BYTES
```
sifnoc, line 32: In this context, does `a` refer to `lhs` and `b` refer to `rhs`?

zk_prover/src/chips/less_than_check/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+    };
+
+    // lhs = 1
+    // rhs = 200
+    // N_BYTES = 2
+    #[test]
+    fn valid_check_lt() {
+        let k = 9;
+
+        // a: new value
```
sifnoc, line 130: Also, In this context, does `a` refer to `lhs`? or is it just left over?"

sifnoc (changes requested):
That's good! There are valuable insights to be gained by patterning after the Axiom one.

enricobottazzi:
@sifnoc Thanks for calling out this typos!! I fixed them and will merge as soon as the tests pass

PR #141, opened 2023-08-28
Refactor smart contract

alxkzmn:
## Overview

The new version of the Summa Smart Contract introduces changes aimed at addressing limitations in querying non-EVM balances and in verifying address ownership across different chains.

### Added

1. **New Structs:**
    - `AddressOwnershipProof` replaces `OwnedAddress` for representing proof of address ownership.
    - `Asset` replaces `OwnedAsset` for representing assets.
    - `ZKProof` introduced to bundle zero-knowledge proof and its inputs.

2. **Event Changes:**
    - Added `ProofOfSolvencySubmitted` event to include asset information.
    - Added `ExchangeAddressesSubmitted` event for capturing multiple address ownership proofs.

3. **Public Mappings and Arrays:**
    - `addressOwnershipProofs` array for tracking all address ownership proofs.
    - `proofOfSolvency` mapping for storing proofs of solvency by timestamp.

4. **Utility Functions:**
    - Added `getProofOfSolvency` to retrieve stored proofs of solvency by timestamp.

### Removed

1. **Address and Balance Retrievers:**
    - Removed `IBalanceRetriever` and `IAddressOwnershipVerifier` interfaces, and corresponding mappings and functions (`setBalanceRetriever`, `setAddressOwnershipVerifier`).

2. **Owned Address and Asset Structs:**
    - Removed `OwnedAddress` and `OwnedAsset` structs.

3. **Event Changes:**
    - Removed `AddressVerifierSet` and `BalanceRetrieverSet` events.

### Modified

1. **Validation Checks:**
    - Optimistic validation for future verification of asset balances and addresses.

2. **Proof Submissions:**
    - Updated `submitProofOfAddressOwnership` and `submitProofOfSolvency` methods to accommodate new structs and logic.

### Removed: Asset Checking from ZK-Proof:

- Address ownership proofs are submitted without verification, as querying of non-EVM chains is currently not feasible.


enricobottazzi:
Meeting here => https://meet.google.com/yum-kswd-rbh

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
-        bytes ownershipProof;
+    struct AddressOwnershipProof {
+        string cexAddress;
+        string chain;
+        bytes signature;
+        bytes message;
     }
 
     /**
      * @dev Struct representing an asset owned by the CEX
```
enricobottazzi, line 29: `amount` missing from comment

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+     */
+    struct ZKProof {
+        bytes proof;
+        uint256[] publicInputs;
+    }
+
+    IVerifier private immutable verifier;
+
+    //All address ownership proofs submitted by the CEX
+    AddressOwnershipProof[] public addressOwnershipProofs;
```
enricobottazzi, line 52: Both the address ownership proofs and the proofs of solvency are:
1. Written to the smart contracts => `addressOwnershipProofs` and `proofOfSolvency`
2. Emitted in events `ExchangeAddressesSubmitted` and `ProofOfSolvencySubmitted`

The goal of the contract is to store (or log) these information such that external actors can challenge it if malicious. 

I'm wondering if you see a difference between writing these to the smart contract and emitting them through events or is it just redundancy.

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+
+    //Convenience mapping to check if an address has already been verified
+    mapping(bytes32 => AddressOwnershipProof) public ownershipProofByAddress;
```
sifnoc, line 55: How about using `i`, index of `addressOwnershipProofs` instead of `AddressOwnershipProof` in here?

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+
+    //All proofs of solvency by timestamp
+    mapping(uint256 => ZKProof) public proofOfSolvency;
```
enricobottazzi, line 58: rename to `solvencyProofs`

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+
+    event ExchangeAddressesSubmitted(
```
enricobottazzi, line 60: rename to `AddressOwnershipProofSubmitted`

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
+        AddressOwnershipProof[] addressOwnershipProofs
+    );
+    event ProofOfSolvencySubmitted(
```
enricobottazzi, line 63: rename to `SolvencyProofSubmitted`

enricobottazzi (commented):
LGTM after these small changes! I love PRs that drastically reduce the code base ♥️

contracts/src/Summa.sol
```diff
@@ -9,163 +9,128 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
     /**
      * @dev Submit proof of solvency for a CEX
-     * @param assets Assets owned by CEX
-     * @param mstRoot The root of the Merkle sum tree
+     * @param mstRoot Merkle sum tree root of the CEX's liabilities
+     * @param assets The list of assets owned by the CEX
      * @param proof The ZK proof
      * @param timestamp The timestamp at which the CEX took the snapshot of its liabilites
      */
     function submitProofOfSolvency(
```
sifnoc, line 109: I noticed that the submitProofOfSolvency function has been updated to align with an optimistic approach. However, the optimistic approach has not yet been developed for Proof of Ownership.
So, I think it would be helpful to add short notification about the optimistic process that is still a work in progress

sifnoc (changes requested):

contracts/src/Summa.sol
alxkzmn: I wanted to use the address (or its hash) for convenience. First, there's a check whether this particular address has previously been verified (line 84), so the check should perform some kind of lookup by address. Second, even in case of manual querying by a human, they would probably be searching for the proof of ownership of a particular address, so the mapping allows to do that in a single query (instead of iterating over the array).

alxkzmn: Actually a good idea! Will change.

contracts/src/Summa.sol
alxkzmn: Done!

contracts/src/Summa.sol
enricobottazzi: - I think that the process of signing proof might possibly added in the future but is not in plan now. Maybe it will be added after the discussion with exchanges 
- If we don't need to store the proofOfSolvency I think this can be safely removed.

contracts/src/Summa.sol
enricobottazzi: What do you mean "still work in progress"?

enricobottazzi (changes requested):
Some tests are failing

alxkzmn:
Seems to be a GitHub glitch

enricobottazzi (approved):

contracts/src/Summa.sol
sifnoc: I though we have a plan to develop the optimistic approach for Proof of ownership. 
Technically it is not "work in progress".

sifnoc (approved):

Issue #147, opened 2023-08-29
Add `ProofOfInclusion` Verifier view function inside the contract 


enricobottazzi:
It should take as input, proof, `username` and `balances[]` and the public inputs of the proof, namely `leaf_hash` and `root_hash`.

- Should verify the proof 
- Should verify that `H(username, balances[])` match `leaf_hash`
- Should verify that `root_hash` matches the one committed by the Exchange in `SubmitProofOfSolvency`

EDIT: maybe we should drop the second functionality and not pass `username` and `balances[]` as input at all. Even tough the function is not executed on chain, the user will likely use a third party node to perform such operation and therefore might not want to share these personal data

alxkzmn:
I think it's better to keep the username and balances private.

PR #149, opened 2023-09-01
Add inclusion verifier to the smart contract

alxkzmn:
See #147


contracts/src/Summa.sol
```diff
@@ -47,14 +47,19 @@ contract Summa is Ownable {
...
+    IVerifier private immutable solvencyVerifier;
+    IVerifier private immutable inclusionVerifier;
 
     //All address ownership proofs submitted by the CEX
     AddressOwnershipProof[] public addressOwnershipProofs;
 
     //Convenience mapping to check if an address has already been verified
     mapping(bytes32 => uint256) public ownershipProofByAddress;
 
+    // MST roots correasponding to successfully verified solvency proofs by timestamp
```
enricobottazzi, line 60: typo

zk_prover/examples/gen_inclusion_verifier.rs
```diff
@@ -0,0 +1,71 @@
...
+        },
+    },
+    merkle_sum_tree::MerkleSumTree,
+};
+
+const LEVELS: usize = 4;
+const N_ASSETS: usize = 2;
+const N_BYTES: usize = 14;
+
+fn main() {
```
enricobottazzi, line 24: Now that I see this function again, I realize that it might be a bit confusing since to generate the verifier you don't really need to instantiate the circuit with a `csv` input but you can use the `init_empty()` API.

Conversely, you need the `csv` file input when generating the test proof.

I'll open a separed issue for this

enricobottazzi (approved):

contracts/test/Summa.ts
```diff
@@ -346,9 +386,223 @@ describe("Summa Contract", () => {
...
+      ).to.be.revertedWith("Invalid MST root");
+    });
+
+    it("should not verify if the MST root lookup by timestamp returns an incorrect MST root", async () => {
+      // The lookup will return a zero MST root as no MST root has been stored yet
+      await expect(
+        summa.verifyInclusionProof(
+          inclusionProof,
+          [leafHash, mstRoot],
+          1693559255
```
sifnoc, line 543: If we use a different timestamp here, for example `1693559256`, we could move the code lines that generate transactions with `await summa.submitProofOfAddressOwnership` and `await summa.submitProofOfSolvency` to a `beforeEach` block. 

This would streamline the setup for all inclusion tests.

sifnoc (approved):
I've left a few minor suggestions, but aside from those, everything looks good to me.

contracts/test/Summa.ts
alxkzmn: I get the idea of extracting the functions but I don't get how it's related to a particular value of the timestamp.

sifnoc: What I meant is that querying `mstRoots` with a timestamp that hasn't been set will return 0 for the `mstRoot`. 
However, it looks like you've already taken care of that in the code, so we're good to go.

Issue #145, opened 2023-08-29
Remove fetch API


enricobottazzi:
The latest update of summa, after #141 is merged, will support any cryptocurrencies with no difference or preference. 

Therefore, the fetch API should be able to fetch balances from any blockchain at any block. The current backend implementation only partially supports it => now only able to fetch ETH (at any block) and ERC20 (only at the most recent block).

I believe that removing the feature as a whole is better than having a partially functioning feature.

sifnoc:
This issue can be resolved following #152
The related commit can be found [9a93558](https://github.com/summa-dev/summa-solvency/pull/152/commits/9a935586f36e1d3ba11c04bc828bf0bb1468c3d6)

Issue #128, opened 2023-08-03
Add whole flow example to backend examples


enricobottazzi:
It would be cool to add the whole flow to `backend/examples`

It should involve: 

- Message signatures to build `csv` file.
- Snapshot generation 
- Proof of Address ownership generation and submission to smart contract 
- Proof of Solvency generation and submission to smart contract 
- Proof of Mst Inclusion generation 
- Proof of Mst Inclusion verification (user side)

Ideally, the example should contain as few lines of code as possible, as all these actions should be abstracted by the backend APIs. This exercise can also inform a better design of the APIs if needed.

enricobottazzi:
@sifnoc after #152 is merged

I guess that according to the modification we can provide in the example both 

- Proof of Mst Inclusion Verification (user side - performed locally using the Verification Key)
- Proof of Mst Inclusion Verification (user side - performed interacting with the view function of the Summa Smart Contract)

sifnoc:
Related `Message signatures to build csv file.` task, Ensure that the Ethereum addresses in the file adhere to the checksum format

enricobottazzi:
https://github.com/summa-dev/summa-solvency/pull/152#discussion_r1319703249

I think that the new example flow will just need to interact with `AddressOwnership` and `Round` objects. Therefore we can make the `Snapshot` struct as private.

sifnoc:
- [x] Message signatures to build `csv` file.
- [ ] ~~Snapshot generation~~
- [x] Proof of Address ownership generation and submission to smart contract 
- [x] Proof of Solvency generation and submission to smart contract
- [ ] Proof of Mst Inclusion verification (user side - locally using given Verification Key)
- [ ] Proof of Mst Inclusion Verification (user side - performed interacting with the view function of the Summa Smart Contract)

PR #152, opened 2023-09-06
Refactor backend following updated contracts

sifnoc:
Update the backend API to reflect changes in the contracts.

Major changes include separating `AddressOwnership` from `Snapshot`, and introducing a new struct, `Round`, which is now included within `Snapshot`.

Modified `gen_inclusion_verifier` and `gen_solvency_verifier` scripts with using specific `ptau` file instead of using random.


zk_prover/examples/gen_solvency_verifier.rs
```diff
@@ -50,6 +50,12 @@ fn main() {
 
     let circuit = SolvencyCircuit::<N_ASSETS, N_BYTES>::init(merkle_sum_tree, asset_sums);
 
+    // generate a universal trusted setup for testing, along with the verification key (vk) and the proving key (pk).
```
enricobottazzi, line 53: Can you mention it in the readme ? both for `gen_solvency_verifier` and `gen_inclusion_verifier`

backend/src/apis/ownership.rs
```diff
@@ -0,0 +1,37 @@
...
+use ethers::types::Address;
+use std::{error::Error, result::Result};
+
+pub struct AddressOwnership {
+    address_ownership_proofs: Vec<AddressOwnershipProof>,
+    signer: SummaSigner,
+}
+
+impl AddressOwnership {
+    pub fn new(
```
enricobottazzi, line 11: I propose that the `new` function takes a csv file containing the triples {address, signature, message} as input, then build `address_ownership_proofs` and returns a struct that already has the proofs in it

backend/src/apis/ownership.rs
```diff
@@ -0,0 +1,37 @@
...
+        rpc_url: &str,
+        address: Address,
+    ) -> Result<AddressOwnership, Box<dyn Error>> {
+        Ok(AddressOwnership {
+            address_ownership_proofs: vec![],
+            signer: SummaSigner::new(&[], main_signer_key, chain_id, rpc_url, address),
+        })
+    }
+
+    // Make sure the input `addresses` in `address_ownership_proof` are not duplicated with addresses already registered on the contract.
```
enricobottazzi, line 23: For what I can tell, this operation of making sure that the input 'addresses` ... is not being performed now so the comment is misleading. BTW I agree that we shouldn't do that.

backend/src/apis/ownership.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub async fn dispatch_proof_of_address_ownership(
+        &mut self,
+        address_ownership_proof: Vec<AddressOwnershipProof>,
```
enricobottazzi, line 26: Based on the current implementation, we expect the Exchange to build these `AddressOwnershipProof` by themselves without providing any helpers for doing that.

So, according to the previous suggestion, the function doesn't have to take address_ownership_proof as input anymore as there are already created in `new`.

Therefore, the only role of this function is to dispatch the proof to the contract

backend/src/apis/ownership.rs
```diff
@@ -0,0 +1,37 @@
...
+    address_ownership_proofs: Vec<AddressOwnershipProof>,
+    signer: SummaSigner,
+}
+
+impl AddressOwnership {
+    pub fn new(
+        main_signer_key: &str,
+        chain_id: u64,
+        rpc_url: &str,
+        address: Address,
```
enricobottazzi, line 15: name it `summa_sc_address`

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+    signer: SummaSigner,
+}
+
+impl<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize>
+    Round<LEVELS, N_ASSETS, N_BYTES>
+where
+    [usize; N_ASSETS + 1]: Sized,
+    [usize; 2 * (1 + N_ASSETS)]: Sized,
+{
+    pub fn new(
```
enricobottazzi, line 79: Add as input to the `new` function
- a csv file related to the liabilities state 
- a csv file related to the asset state
- timestamp related to the round in UNIX format
- params path

Therefore the role of the function will be to
build the Snapshot by calling `build_snapshot`, together with the Signer, and attach it to the Round object

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+        signer_key: &str,
+        chain_id: u64,
+        rpc_url: &str,
+        summasc_address: Address,
```
enricobottazzi, line 83: rename to `summa_sc_address`

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub fn get_public_inputs(&self) -> &Vec<Vec<Fp>> {
+        &self.public_inputs
+    }
+
+    pub fn get_proof(&self) -> &Vec<u8> {
+        &self.proof
+    }
+}
+
+pub struct Snapshot<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize> {
```
enricobottazzi, line 62: Add `assets_state` to the Snapshot, it will containt `[Asset; N_ASSETS]`

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+        rpc_url: &str,
+        summasc_address: Address,
+    ) -> Result<Round<LEVELS, N_ASSETS, N_BYTES>, Box<dyn Error>> {
+        Ok(Round {
+            snapshot: None,
+            signer: SummaSigner::new(&vec![], signer_key, chain_id, rpc_url, summasc_address),
+        })
+    }
+
+    pub fn build_snapshot(&mut self, entry_csv_path: &str, params_path: &str, timestamp: usize) {
```
enricobottazzi, line 91: I think this function is not needed anymore as you can perform it directly in `new`

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+        let snapshot =
+            Snapshot::<LEVELS, N_ASSETS, N_BYTES>::new(entry_csv_path, params_path, timestamp)
+                .unwrap();
+        self.snapshot = Some(snapshot);
+    }
+
+    pub async fn dispatch_solvency_proof(
```
enricobottazzi, line 98: According to the previous comments`dispatch_solvency_proof` should take no inputs

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+}
+
+impl<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize>
+    Snapshot<LEVELS, N_ASSETS, N_BYTES>
+where
+    [usize; N_ASSETS + 1]: Sized,
+    [usize; 2 * (1 + N_ASSETS)]: Sized,
+{
+    pub fn new(
```
enricobottazzi, line 152: `timestamp` is not used here. We can keep `timestamp` inside the Round, it doesn't need to be used inside Snapshot

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+        ];
+
+        Ok(Snapshot {
+            mst,
+            timestamp,
+            trusted_setup,
+        })
+    }
+
+    pub fn get_timestamp(&self) -> usize {
```
enricobottazzi, line 185: `get_timestamp` method can be moved to Round

backend/src/apis/fetch.rs
```diff
@@ -54,7 +54,7 @@ mod tests {
     use super::*;
 
     use crate::contracts::mock::mock_erc20::MockERC20;
-    use crate::contracts::tests::initialize_anvil;
+    use crate::tests::initialize_anvil;
```
enricobottazzi, line 57: I think we can remove the `fetch` API as mentioned in issue [#145](https://github.com/summa-dev/summa-solvency/issues/145). Also all the related tests

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+}
+
+pub struct Snapshot<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize> {
+    mst: MerkleSumTree<N_ASSETS, N_BYTES>,
+    timestamp: usize,
+    trusted_setup: [SetupArtifacts; 2],
+}
+
+pub struct Round<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize> {
```
enricobottazzi, line 68: Add `timestamp` to Round

backend/src/apis/round.rs
```diff
@@ -0,0 +1,226 @@
...
+        let snapshot = self.snapshot.as_ref().unwrap();
+        if snapshot.mst.entries().len() < user_index {
+            return Err("user_index is out of range");
+        }
+
+        Ok(snapshot.generate_proof_of_inclusion(user_index).unwrap())
+    }
+}
+
+impl<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize>
```
enricobottazzi, line 146: I believe that Snapshot doesn't have to be exposed to the public since the exchange would only potentially interact with `Round`, therefore I think that we can safely remove `pub` from these function. I'm not sure about that tho

enricobottazzi (changes requested):

backend/src/apis/ownership.rs
alxkzmn, line 1: Let's rename the file as address_ownership

backend/src/tests.rs
```diff
@@ -170,11 +139,20 @@ mod test {
             ("0xb17a9e25265d3b88de7bfad81e7accad6e3d5612308ff83cc0fef76a34152b0444309e8fc3dea5139e49b6fc83a8553071a7af3d0cfd3fb8c1aea2a4c171729c1c").parse().unwrap(),
             message:  "Summa proof of solvency for CryptoExchange".encode().into(),
         }];
-        let result = summa_signer
-            .submit_proof_of_address_ownership(owned_addresses)
+
+        let mut address_ownership_client = AddressOwnership::new(
```
alxkzmn, line 143: I think that the address ownership should be a part of the Round (similarly to Snapshot). While it's possible to submit it at any time before the proof, it's still enough to submit it in the beginning of a round. Most likely it wouldn't be called manually in the production implementation of Summa like you're calling it here. There should be some entity that controls it and Round seems to be a good candidate for that.

alxkzmn (changes requested):

backend/src/apis/ownership.rs
alxkzmn: Good point, that's why I also think that the address ownership should live inside the Round and Round should take care of it. (cc @sifnoc )

alxkzmn: I'm not sure if they'll be performed separately in the real-world implementation. E.g., CEX can add more addresses in between the proving rounds but in each round it has to check if all the addresses have the ownership proofs submitted, so it makes sense just to submit new addresses in the beginning of the round. BTW I agree that it's good to separate address ownership into its own trait.

backend/src/apis/round.rs
sifnoc: Currently, the snapshot is directly used in the inclusion verifier example, which is why I've made it public. On the issue #128 I'll consider making it private if it's appropriate.

backend/src/apis/ownership.rs
sifnoc: Updated at [ce908f0](https://github.com/summa-dev/summa-solvency/pull/152/commits/ce908f0a7323153aea8f3f959f16e69766a661d7)

backend/src/apis/ownership.rs
sifnoc: Updated at [ce908f0](https://github.com/summa-dev/summa-solvency/pull/152/commits/ce908f0a7323153aea8f3f959f16e69766a661d7)

backend/src/apis/ownership.rs
sifnoc: Updated comment at [ce908f0](https://github.com/summa-dev/summa-solvency/pull/152/commits/ce908f0a7323153aea8f3f959f16e69766a661d7)

Both perspectives are valid. However, from my understanding, the `AddressOwnership` operation isn't intrinsically tied to `Round`. As such, it might not be fitting to merge `AddressOwnership` directly into `Round` at this juncture.
That being said, there could be value in creating a new struct that encompasses both functionalities. 
Perhaps something like `SummaOperator` or `SummaController`?
For a CEX operator, a consolidated interface for both `Round` and `AddressOwnership` would undoubtedly be beneficial.

backend/src/apis/ownership.rs
sifnoc: Updated comment at [ce908f0](https://github.com/summa-dev/summa-solvency/pull/152/commits/ce908f0a7323153aea8f3f959f16e69766a661d7)

backend/src/apis/ownership.rs
sifnoc: Updated comment at [ce908f0](https://github.com/summa-dev/summa-solvency/pull/152/commits/ce908f0a7323153aea8f3f959f16e69766a661d7)

backend/src/apis/round.rs
sifnoc: Updated comment at [ce908f0](https://github.com/summa-dev/summa-solvency/pull/152/commits/ce908f0a7323153aea8f3f959f16e69766a661d7)

backend/src/apis/fetch.rs
sifnoc: Removed [3d1ab95](https://github.com/summa-dev/summa-solvency/pull/152/commits/3d1ab95db73fcfe38006aa7f300b3a78f15fd678)

backend/src/apis/round.rs
sifnoc: Updated at [3d1ab95](https://github.com/summa-dev/summa-solvency/pull/152/commits/3d1ab95db73fcfe38006aa7f300b3a78f15fd678)

backend/src/apis/round.rs
sifnoc: Updated at [3d1ab95](https://github.com/summa-dev/summa-solvency/pull/152/commits/3d1ab95db73fcfe38006aa7f300b3a78f15fd678)

backend/src/apis/round.rs
sifnoc: Updated at [3d1ab95](https://github.com/summa-dev/summa-solvency/pull/152/commits/3d1ab95db73fcfe38006aa7f300b3a78f15fd678)

backend/src/apis/round.rs
sifnoc: I've developed a CSV parser for the asset state, which directly generates data types, `Assets` as defined in the contract interface.
[`3d1ab95` (#152)](https://github.com/summa-dev/summa-solvency/pull/152/commits/3d1ab95db73fcfe38006aa7f300b3a78f15fd678#diff-5af8cf7679db328255bf2c9ffd6420b71f1c2a97d407f86d9259c0fca8a17f0a)

backend/src/apis/round.rs
sifnoc: Updated at [3d1ab95](https://github.com/summa-dev/summa-solvency/pull/152/commits/3d1ab95db73fcfe38006aa7f300b3a78f15fd678)

backend/src/tests.rs
sifnoc: https://github.com/summa-dev/summa-solvency/pull/152#discussion_r1321759814 partially addresses your suggestion for now.
A potential alternative could be the `Scheduler` you introduced in the [Service-Based Backend Architecture](https://hackmd.io/CXoZfrc0Ri-UOd-whxvvnA)
Let's discuss this further in the document.

backend/src/apis/csv_parser.rs
```diff
@@ -1,49 +1,118 @@
...
+    abi::AbiEncode,
+    types::{Bytes, U256},
+};
 use serde::Deserialize;
 
+use crate::contracts::generated::summa_contract::{AddressOwnershipProof, Asset};
+
 #[derive(Debug, Deserialize)]
-struct Record {
+struct EntriesRecord {
```
enricobottazzi, line 12: Can we rename it to `SignatureRecord` as the other is named `AssetRecord` ?

enricobottazzi (approved):
Just had a single comment on a variable name. LGTM!

backend/src/apis/csv_parser.rs
```diff
@@ -1,49 +1,118 @@
...
+            asset_name: record.asset_name,
+            chain: record.chain,
+            amount: U256::from_dec_str(&record.amount)?,
+        });
     }
 
-    Ok((addresses, signatures))
+    let assets_array: [Asset; N_ASSETS] = assets_vec.try_into().map_err(|v: Vec<Asset>| {
+        format!(
+            "Number of assets in CSV does not match the number of assets in the contract! {:?}",
```
alxkzmn, line 68: The error message is misleading, it should say something like "the number of assets doesn't match the expected"

alxkzmn (approved):

backend/src/apis/ownership.rs
alxkzmn: I'm not talking about merging (similarly to how Snapshot can still live in a separate file and not inside the same file with Round). I'm only talking about the encapsulation of AddressOwnership inside Round (Round holding an instance of AddressOwnership). As I imagine the future use of the system, CEX will not call any of Summa services from within CEX services, they will be called by the scheduler. If everything gets called with the same interval, it's all a "Round" of a proof of solvency.

alxkzmn (approved):

backend/src/apis/ownership.rs
sifnoc: I now understand your perspective. You're suggesting a `Round` structure similar to this:
```Rust
pub struct Round<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize> {
    timestamp: u64,
    snapshot: Snapshot<LEVELS, N_ASSETS, N_BYTES>,
    address_ownership: AddressOwnership,
    signer: SummaSigner,
}
```

Upon reviewing the following code segment, I see its alignment with your viewpoint on AddressOwnership:
https://github.com/summa-dev/summa-solvency/blob/c4edbc6fe75131690c9f323be333e2d5c12e20df/contracts/src/Summa.sol#L107-L110

However, regardless of the naming convention for `Round`, I believe there should be only one `Signer`. 
If there are multiple signers, they might generate transactions with the same nonce.

enricobottazzi: > I'm only talking about the encapsulation of AddressOwnership inside Round (Round holding an instance of AddressOwnership). As I imagine the future use of the system, CEX will not call any of Summa services from within CEX services, they will be called by the scheduler. If everything gets called with the same interval, it's all a "Round" of a proof of solvency.

My hypothesis is that `Round` and `AddressOwnership` are two very separate things and the Scheduler will only have to deal with `Round`. The soundness of this hypothesis depends on the frequency of address rotation performed by an Exchange. If they stick to the same address for long time, `AddressOwnership` is just called once and then just updated once in a while. On the contrary, if the switch addresses very frequently, it's likely that at every round a new proof of address ownership must be called.

I think we can stick to this now and challenge this hypothesis as soon as speak with customers. This will also inform the design of the Summa services.

enricobottazzi: > However, regardless of the naming convention for `Round`, I believe there should be only one `Signer`.
> If there are multiple signers, they might generate transactions with the same nonce.

Can elaborate on that @sifnoc ?

sifnoc: In the current implementation, both `Round` and `AddressOwnership` have an instance of Signer, each. 
The default middleware in `ethers-rs` does not provide a shared nonce counter between signers. 
As a result, both signers might query the node using `get_transaction_count` for the same `Account`. 

This can lead to a high likelihood that both signers will generate transactions with the same nonce for `submit_proof_of_solvency` and `submit_proof_of_address_ownership`. 

Consequently, one of these transactions may not be included in the block.

alxkzmn: Yes, the same Signer should be injected everywhere where needed.

backend/src/apis/csv_parser.rs
sifnoc: Updated at [31c3c59](https://github.com/summa-dev/summa-solvency/pull/152/commits/31c3c59bf0d5efa6027b238ee7bb54bceb1b8800)

backend/src/apis/csv_parser.rs
sifnoc: Updated at [31c3c59](https://github.com/summa-dev/summa-solvency/pull/152/commits/31c3c59bf0d5efa6027b238ee7bb54bceb1b8800)

PR #156, opened 2023-09-14
Add MST leaf update function

alxkzmn:
Added a method to update a designated leaf of the MST. Added the capability to build an MST with the leaves sorted by the username value. The motivation for sorting is the following. To update an entry, you need to know its index. In our implementation, each entry has a unique username, and the balance updates should happen for specific users. Since the users are unique, it's convenient to sort the entries by username to reduce the user search time from linear to logarithmic in the size of the MST leaf level. Added a benchmark for the sorted MST - the sorting overhead seems to be negligible.


zk_prover/src/merkle_sum_tree/entry.rs
```diff
@@ -32,22 +32,28 @@ impl<const N_ASSETS: usize> Entry<N_ASSETS> {
     }
 
     pub fn compute_leaf(&self) -> Node<N_ASSETS>
+    where
+        [usize; N_ASSETS + 1]: Sized,
+    {
+        self.create_node(&self.balances)
```
enricobottazzi, line 38: As you added the api `middle` on the struct `Node`, can you also add an api `leaf` that replaces the function `create_node` on `Entry`?

zk_prover/src/merkle_sum_tree/tests.rs
```diff
@@ -87,6 +87,85 @@ mod test {
         proof_invalid_3.sibling_sums[0] = [0.into(), 0.into()];
     }
 
+    #[test]
+    fn test_update_mst_leaf() {
```
enricobottazzi, line 91: Can you create a test (or modify this one) that: 

- Build a first merkle tree from `entry_16`
- Update leaf 7 of the merkle tree with balances 2087, 79731
- Build a second merkle tree from `entry_16_updated` that has the 7th username with balances 2087, 79731
- Check that the merkle root and the merkle balances of the first updated tree and the second tree are matching

enricobottazzi (changes requested):

zk_prover/src/merkle_sum_tree/mst.rs
```diff
@@ -86,12 +158,20 @@ impl<const N_ASSETS: usize, const N_BYTES: usize> MerkleSumTree<N_ASSETS, N_BYTE
         Ok((&penultimate_level[0], &penultimate_level[1]))
     }
 
-    /// Returns the index of the user with the given username and balances in the tree
+    /// If the tree is unsorted, returns the index of the leaf with the matching username and balances. If the tree is sorted (by username), returns the index of the leaf with the matching username.
```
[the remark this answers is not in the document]
alxkzmn: Binary search in a sorted array is logarithmic, simple search in an unsorted array is linear.

zk_prover/src/merkle_sum_tree/mst.rs
```diff
@@ -38,6 +39,34 @@ impl<const N_ASSETS: usize, const N_BYTES: usize> MerkleSumTree<N_ASSETS, N_BYTE
         [usize; 2 * (1 + N_ASSETS)]: Sized,
     {
         let entries = parse_csv_to_entries::<&str, N_ASSETS, N_BYTES>(path)?;
+        Self::build_tree(entries, false)
+    }
+
+    /// Builds a Merkle Sum Tree from a CSV file stored at `path`. The MST leaves are sorted by the username value. The CSV file must be formatted as follows:
```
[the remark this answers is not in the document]
alxkzmn: It's generally irrelevant how you sort when you sort it for the sole purpose of quick binary search. As it turns out from the docs, the string sorting is implemented in Rust by the byte values, so it's not always alphabetical (e.g., for Unicode, but may be alphabetical if we are sure that the hashed usernames only contain ASCII).

alxkzmn:
Updated the description.

enricobottazzi (approved):

zk_prover/src/merkle_sum_tree/mst.rs
```diff
@@ -55,9 +84,50 @@ impl<const N_ASSETS: usize, const N_BYTES: usize> MerkleSumTree<N_ASSETS, N_BYTE
...
+    ///
+    /// # Arguments
+    ///
+    /// * `username`: The username of the entry to update
+    /// * `new_balances`: The new balances of the entry
+    ///
+    /// # Returns
+    ///
+    /// The new root of the tree
+    pub fn update_leaf(
```
sifnoc, line 101: Using `update_leaf` as a way to avoid rebuilding the entire Merkle tree each time seems efficient. However, I suspect that updating the tree with `update_leaf` might be slower than `build_tree` when more than half of the user balances are being modified, especially since `build_tree` employs a multi-threaded approach. 
Do you have any references or benchmarks regarding this?

zk_prover/src/merkle_sum_tree/mst.rs
```diff
@@ -38,6 +39,34 @@ impl<const N_ASSETS: usize, const N_BYTES: usize> MerkleSumTree<N_ASSETS, N_BYTE
...
         let entries = parse_csv_to_entries::<&str, N_ASSETS, N_BYTES>(path)?;
+        Self::build_tree(entries, false)
+    }
+
+    /// Builds a Merkle Sum Tree from a CSV file stored at `path`. The MST leaves are sorted by the username byte values. The CSV file must be formatted as follows:
+    ///
+    /// `username;balances`
+    ///
+    /// `dxGaEAii;11888,41163`
+    pub fn new_sorted(path: &str) -> Result<Self, Box<dyn std::error::Error>>
```
sifnoc, line 50: To ensure my understanding is correct, let's assume we have a sorted MST in round 1. 
If a new user is added in round 2, would we need to reconstruct the entire MST?

zk_prover/src/merkle_sum_tree/mst.rs
alxkzmn: Points from discussion at the meeting
- parallelized batch insertion can be a good compromise between a single leaf update and a tree rebuild;
- the relation between the user balance update frequency and the proof solvency round frequency would define the ratio of updated leaves but it is not known yet;
- we can benchmark if we modify the test data to contain only unique usernames.

zk_prover/src/merkle_sum_tree/mst.rs
alxkzmn: Yes.

sifnoc (approved):
Thanks for updating :)

PR #164, opened 2023-10-23
Fix OOM while building MST

sifnoc:
There is OOM when build Merkle some tree that has more than 2^27 elements due to no limit on the `thread`.
We can apply `Rayon` for multi-threading instead of `thread`.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -2972,9 +2972,9 @@ dependencies = [
 
 [[package]]
 name = "rayon"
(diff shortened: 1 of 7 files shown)


enricobottazzi:
LGTM! Can you please just update the `README` with the new benchmarks results (I noticed you modified from 2 assets to 1 assets)

enricobottazzi (approved):

alxkzmn (approved):

PR #168, opened 2023-10-31
chore: add audit comments

sifnoc:


contracts/src/Summa.sol
```diff
@@ -77,6 +89,9 @@ contract Summa is Ownable {
             );
             uint256 index = ownershipProofByAddress[addressHash];
             require(index == 0, "Address already verified");
+            /*
+              Is there any reason to assign value `i + 1` to `ownershipProofByAddress[addressHash]`?
```
alxkzmn, line 93: The idea was to store the array index, but starting with 1 allows to distinguish a case when the proof has not yet been submitted (the storage slots are zero by default).

PR #174, opened 2023-11-01
Separated signer from core components.

sifnoc:
- `AddressOwnership` and `Round` now require a `Signer` for initialization. and they use same signer instance.
- The `Signer` has a mutex lock for preventing nonce collision.
- Added test case that submit two solvency proofs at the same time.

### DIFF
--- a/backend/examples/summa_solvency_flow.rs
+++ b/backend/examples/summa_solvency_flow.rs
@@ -20,7 +20,7 @@ const USER_INDEX: usize = 0;
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     // Initialize test environment without `address_ownership` instance from `initialize_test_env` function.
-    let (anvil, _, _, _, summa_contract) = initialize_test_env().await;
+    let (anvil, _, _, _, summa_contract) = initialize_test_env(None).await;
 
     // 1. Submit ownership proof
     //
--- a/backend/examples/summa_solvency_flow.rs
+++ b/backend/examples/summa_solvency_flow.rs
@@ -1,7 +1,7 @@
 #![feature(generic_const_exprs)]
-use std::{error::Error, fs::File, io::BufReader, io::Write, sync::Arc};
+use std::{error::Error, fs::File, io::BufReader, io::Write};
 
-use ethers::{providers::Provider, types::U256};
+use ethers::types::U256;
 use serde_json::{from_reader, to_string_pretty};
 
 use summa_backend::{
@@ -27,18 +27,17 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Each CEX prepares its own `signature` CSV file.
     let signature_csv_path = "src/apis/csv/signatures.csv";
 
-    // The signer would be using `provider` that shared with `address_ownership` and `round` instances.
-    let provider = Arc::new(Provider::try_from(anvil.endpoint().as_str())?);
-
+    // The signer instance would be shared with `address_ownership` and `round` instances
+    //
     // Using AddressInput::Address to directly provide the summa_contract's address.
     // For deployed contracts, if the address is stored in a config file,
     // you can alternatively use AddressInput::Path to specify the file's path.
     let signer = SummaSigner::new(
         "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
-        anvil.chain_id(),
-        provider,
+        anvil.endpoint().as_str(),
         AddressInput::Address(summa_contract.address()),
-    );
+    )
+    .await?;
 
     let mut address_ownership_client = AddressOwnership::new(&signer, signature_csv_path).unwrap();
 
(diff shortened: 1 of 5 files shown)


backend/src/apis/address_ownership.rs
```diff
@@ -1,29 +1,23 @@
...
 use std::{error::Error, result::Result};
 
 use super::csv_parser::parse_signature_csv;
 
-pub struct AddressOwnership {
+pub struct AddressOwnership<'a> {
     address_ownership_proofs: Vec<AddressOwnershipProof>,
-    signer: SummaSigner,
+    signer: &'a SummaSigner,
 }
```
enricobottazzi, line 9: I understand (and appreciate) your choice of using a lifetime annotation for the `SummaSigner` reference. This is the only place in the code base we use that. Do you think there are other places in which this might be needed for managing references?

backend/examples/summa_solvency_flow.rs
```diff
@@ -55,17 +58,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-        anvil.chain_id(),
-        anvil.endpoint().as_str(),
-        AddressInput::Address(summa_contract.address()),
-        entry_csv,
-        asset_csv,
-        params_path,
-        1,
-    )
-    .unwrap();
+    let mut round = Round::<4, 2, 14>::new(&signer, entry_csv, asset_csv, params_path, 1).unwrap();
```
enricobottazzi, line 61: can you create a variable named `timestamp` and assign 1 to that. Passing 1 directly to the `new` function might not be clear for the reader

backend/examples/summa_solvency_flow.rs
```diff
@@ -20,24 +20,27 @@ const USER_INDEX: usize = 0;
...
         anvil.chain_id(),
-        anvil.endpoint().as_str(),
+        provider,
         AddressInput::Address(summa_contract.address()),
-        signature_csv_path,
-    )
-    .unwrap();
+    );
+
+    let mut address_ownership_client = AddressOwnership::new(&signer, signature_csv_path).unwrap();
```
enricobottazzi, line 43: much more clean 😍

backend/src/tests.rs
```diff
@@ -136,18 +170,77 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_concurrent_proof_submissions() -> Result<(), Box<dyn Error>> {
```
enricobottazzi, line 174: Can you make much clear in the commments what is the goal of this test and how the results you are obtaining are proving that the goal was accomplished?

enricobottazzi (approved):
Kudos for thinking about the nonce collision problem and finding an elegant solution for that.

I left some minor comment, but LGTM. 

Relying on @Alex review for checking that the testing is done correctly

backend/examples/summa_solvency_flow.rs
```diff
@@ -20,24 +20,27 @@ const USER_INDEX: usize = 0;
...
+    // The signer would be using `provider` that shared with `address_ownership` and `round` instances.
+    let provider = Arc::new(Provider::try_from(anvil.endpoint().as_str())?);
+
     // Using AddressInput::Address to directly provide the summa_contract's address.
     // For deployed contracts, if the address is stored in a config file,
     // you can alternatively use AddressInput::Path to specify the file's path.
-    let mut address_ownership_client = AddressOwnership::new(
+    let signer = SummaSigner::new(
         "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
         anvil.chain_id(),
```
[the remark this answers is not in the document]
alxkzmn: Simply speaking, the provider url is chain-specific so there shouldn't be a discrepancy between the provider url and chain ID

backend/src/contracts/signer.rs
```diff
@@ -19,29 +18,26 @@ pub enum AddressInput {
 
 #[derive(Debug)]
 pub struct SummaSigner {
-    summa_contract: Summa<SignerMiddleware<Provider<Http>, LocalWallet>>,
+    nonce_lock: Mutex<()>, // To prevent running `submit` methods concurrently
```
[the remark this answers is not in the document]
alxkzmn: Ok, let's consider this resolved.

backend/examples/summa_solvency_flow.rs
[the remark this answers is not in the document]
sifnoc: Do you think it's better to derive the `chain_id` directly from the URL, instead of assigning it explicitly?

backend/src/apis/address_ownership.rs
sifnoc: The lifetime annotation is necessary because the `signer` instance is shared between both `Round` and `AddressOwnership`. I traced the flow of data like this: `Round` → `Snapshot` → `MerkleSumTree` → `Node`, `Entry`. There doesn't seem to be any need for data sharing along that chain.

The only improvement I identified is that there's no need to clone `merkle_sum_tree` when initializing circuits.
For instance, instead of 
`let mut circuit = MstInclusionCircuit::<LEVELS, N_ASSETS, N_BYTES>::init(merkle_sum_tree.clone(), 0);`, 
it could be 
`let mut circuit = MstInclusionCircuit::<LEVELS, N_ASSETS, N_BYTES>::init(&merkle_sum_tree, 0);`.

enricobottazzi: Good point! Gonna create an issue for that

sifnoc: > Good point! Gonna create an issue for that

I got fixed one, let me assign to me :)

backend/examples/summa_solvency_flow.rs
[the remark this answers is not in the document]
alxkzmn: Actually I got into the ethers code and I see that there's an RPC method call inside that retrieves the chain ID from the node: `provider.get_chainid()`. I think with this in mind it's enough to create the provider from the URL and then get the chain ID.

alxkzmn (approved):

Issue #172, opened 2023-10-31
Add `AggregationMerkleSumTree` to `summa-aggregation`


enricobottazzi:
Feature requested as part of "SPECS - "fast and cheap commitment phase". 

According to the feature, a CEX should be able to parallelize the building of a set of "mini" Merkle sum trees across different machines and eventually aggregate them into a bigger Merkle Sum Tree.

Although the functionality involving the parallelization will be applied on the `backend` level, a new data structure `AggregationMerkleSumTree` is required on the `zk-prover` level. This has to:

- Take "mini" Merkle Sum Trees as input and build the `AggregationMerkleSumTree`
- `generate_proof`
- `verify_proof`

enricobottazzi:
According to the discussion we had today, `AggregationTree` will be added to `aggregation-summa` repo implementing the `Tree` trait. 

@sifnoc you can have a look to #177 for the data structure

PR #177, opened 2023-11-01
add `AggregationTree`

enricobottazzi:
#172 

- Create a new `AggregationMerkleSumTree` data structure to `merkle_sum_tree` module
- Add test to generate `MstInclusionProof` starting from `AggregationMerkleSumTree`

### DIFF
--- a/zk_prover/benches/full_solvency_flow.rs
+++ b/zk_prover/benches/full_solvency_flow.rs
@@ -110,8 +110,14 @@ fn generate_zk_proof_mst_inclusion_circuit(_c: &mut Criterion) {
 
     let merkle_sum_tree = MerkleSumTree::<N_ASSETS, N_BYTES>::new(&csv_file).unwrap();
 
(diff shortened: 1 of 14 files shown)


alxkzmn (changes requested):
It may be a good time to refactor the MST-related functions inside the utils and turn them into the MST trait. Now, after you have introduced an aggregate tree, some of the functions serve both data structures (the old "mini" tree and the aggregate tree). It would be nice to have them both implementing an "MST" trait that has the default implementation for shared methods like getting the Merkle proof. This way I could simply write `some_mst.get_proof(&leaf)` - that's what I expect when working with a language that allows for OOP. Having data structures and functions completely separate is not as intuitive - if I'm a new developer coming to the project, I would have to search for the Merkle proof function to find out that it actually lives in a separate file in a separate module.

enricobottazzi:
@alxkzmn great point! gonna apply the changes

PR #178, opened 2023-11-03
Refactor the contract according to V1 consolidation spec

alxkzmn:


contracts/src/Summa.sol
```diff
@@ -37,35 +37,48 @@ contract Summa is Ownable {
...
+     * @param mstRoot Merkle sum tree root of the CEX's liabilities
+     * @param rootSum The total sum of the tree asset
+     */
+    struct Commitment {
+        uint256 mstRoot;
+        uint256 rootSum;
+        Asset asset;
+    }
+
+    //User inclusion proof verifier
```
enricobottazzi, line 51: spacing

contracts/src/Summa.sol
```diff
@@ -77,6 +90,7 @@ contract Summa is Ownable {
             );
             uint256 index = ownershipProofByAddress[addressHash];
             require(index == 0, "Address already verified");
+            //Offsetting the index by 1 to distinguish with the case when the proof hasn't been submitted (the storage slot would be zero)
```
enricobottazzi, line 93: spacing

contracts/test/Summa.ts
```diff
@@ -255,113 +242,85 @@ describe("Summa Contract", () => {
       const calldata: any = JSON.parse(jsonData);
 
       mstRoot = calldata.public_inputs[0];
```
enricobottazzi, line 244: We will no longer have `solvency_proof_solidity_calldata.json` and its generator script in the future. How do you think we should modify the flow accordingly? One suggestion that I have is to modify the `gen_inclusion_verifier.rs` example with `generate_mst_commitment.rs` that outputs something like `commitment_calldata.json`

contracts/test/Summa.ts
```diff
@@ -421,19 +380,20 @@ describe("Summa Contract", () => {
       leafHash = inclusionCalldata.public_inputs[0];
       mstRoot = inclusionCalldata.public_inputs[1];
       inclusionProof = inclusionCalldata.proof;
+      rootSum = BigNumber.from(10000000);
     });
 
     it("should verify the proof of inclusion for the given public input", async () => {
       await summa.submitProofOfAddressOwnership(ownedAddresses);
-      await submitProofOfSolvency(summa, mstRoot, solvencyProof);
+      await submitCommitment(summa, mstRoot, rootSum);
```
enricobottazzi, line 388: I realize that there's a security issue here. You hardcode `rootSum` as 10000000 when sending the commitment. But this is not the correct root sum associated with `mstRoot`. To fix that, the `ProofOfInclusion` should also output the `rootSum` as public output and the verification function should check that it matches the committed one

contracts/test/Summa.ts
```diff
@@ -255,113 +242,85 @@ describe("Summa Contract", () => {
...
-
-      await expect(
-        submitProofOfSolvency(summa, mstRoot, proof)
-      ).to.be.revertedWithoutReason();
-    });
   });
 
   describe("verify proof of inclusion", () => {
     let mstRoot: BigNumber;
+    let rootSum: BigNumber;
```
enricobottazzi, line 323: I cannot comment on the right line but `solvencyProof` should no longer exist here

enricobottazzi (changes requested):

contracts/src/Summa.sol
sifnoc, line 93: Let's illustrate with an example to clarify the behavior of the `submitProofOfAddressOwnership` function. 

In the first transaction, if two proofs are submitted, the `ownershipProofByAddress` mapping will be updated as:
```
{
  hash1: 1,
  hash2: 2
}
```
In a subsequent transaction with three proofs, the mapping will be updated to:
```
{
  hash1: 1,
  hash2: 2,
  hash3: 1,
  hash4: 2,
  hash5: 3
}
```
Is this how the function is supposed to work?

sifnoc (changes requested):

contracts/src/Summa.sol
```diff
@@ -50,34 +60,25 @@ contract Summa is Ownable {
...
 
     event AddressOwnershipProofSubmitted(
         AddressOwnershipProof[] addressOwnershipProofs
     );
     event SolvencyProofSubmitted(
         uint256 indexed timestamp,
         uint256 mstRoot,
-        Asset[] assets
+        uint256 rootSum,
+        Asset assets
```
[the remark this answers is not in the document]
alxkzmn: Because we're now having a single asset per tree per the new specs.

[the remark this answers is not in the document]
enricobottazzi: Roll back to multi-assets as discussed privately

contracts/test/Summa.ts
alxkzmn: I think it's a good idea.

contracts/src/Summa.sol
alxkzmn: Fixed.

contracts/src/Summa.sol
```diff
@@ -94,47 +117,50 @@ contract Summa is Ownable {
...
      * @dev Submit proof of solvency for a CEX
      * @param mstRoot Merkle sum tree root of the CEX's liabilities
-     * @param assets The list of assets owned by the CEX
-     * @param proof The ZK proof
+     * @param rootSums The total sums of the assets included into the Merkle sum tree
+     * @param assets The assets included into the Merkle sum tree
      * @param timestamp The timestamp at which the CEX took the snapshot of its assets and liabilities
      */
-    function submitProofOfSolvency(
+    function submitCommitment(
```
enricobottazzi, line 124: Here we should check that Asset `amount` is greater than `rootSums`

enricobottazzi (changes requested):

contracts/src/Summa.sol
enricobottazzi: Resolved in private discussion

sifnoc (approved):
LGTM!

PR #179, opened 2023-11-06
feat: `Tree` trait

enricobottazzi:
Replacement of #177 

- Added `Tree` trait
- `MerkleSumTree` now implements `Tree` trait
- Modified `build_merkle_tree_from_entries` utils function to `build_merkle_tree_from_leaves` to make it more generic.
- Modified `MerkleProof` struct to take leaf instead of entry. This makes it more generic.
- Modified `MstInclusion` Circuit constructor function to take MerkleProof rather than whole MerkleSumTree

to do @sifnoc :
- [x] `Round` should no longer take a `csv_path` as input. Instead, take an object that implements the `Tree` trait.
- [x] `generate_proof_of_inclusion` API on round should be modified to support the new 
- [x] Update `test` and `example` accordingly
- [x] Import the `summa_solvency::MerkleSumTree` module inside backend module (First, you include the external library in your Cargo.toml file and then use use statement to bring the module into scope in your library code). Then, expose the imported module to other users of `summa_backend` library by using pub use statement. In this way, the `aggregation-tree` repo would only need to use `summa_backend` as dependency while still being able to access the API needed to build the `MerkleSumTree`, which is needed for the workers.


alxkzmn (changes requested):

zk_prover/src/merkle_sum_tree/mst.rs
```diff
@@ -25,9 +26,44 @@ pub struct MerkleSumTree<const N_ASSETS: usize, const N_BYTES: usize> {
...
+        &self.nodes
+    }
+
+    /// Verifies a MerkleProof
+    fn verify_proof(&self, proof: &MerkleProof<N_ASSETS, N_BYTES>) -> bool
+    where
+        [usize; N_ASSETS + 1]: Sized,
+        [usize; 2 * (1 + N_ASSETS)]: Sized,
+    {
+        verify_proof(proof)
```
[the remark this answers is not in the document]
enricobottazzi: I applied the changes in `b3990eb`. I believe that this can only apply to `verify_proof` while I cannot define the logic of `generate_proof` within the `Tree` trait as this will be different between `MerkleSumTree` and `AggregationMerkleSumTree`

alxkzmn (approved):
LGTM!

enricobottazzi:
@sifnoc sorry about that. I think that my comment was cut off. What I meant was to update `generate_proof_of_inclusion` to reflect the new API of the `MstInclusionCircuit` => `https://github.com/summa-dev/summa-solvency/blob/2f14b3633f4fd0dd6ad9a8f75f82807109e9e57d/zk_prover/src/circuits/merkle_sum_tree.rs#L72`

contracts/src/Summa.sol
```diff
@@ -85,6 +85,15 @@ contract Summa is Ownable {
         inclusionVerifier = _inclusionVerifier;
     }
 
+    /*
```
enricobottazzi, line 88: What is this @alxkzmn @sifnoc?

backend/examples/summa_solvency_flow.rs
```diff
@@ -9,35 +9,37 @@ use summa_backend::{
...
     // 1. Submit ownership proof
     //
-    // Each CEX prepares its own `signature` CSV file.
-    let signature_csv_path = "src/apis/csv/signatures.csv";
-
+    // The signer instance would be shared with `address_ownership` and `round` instances
+    //
     // Using AddressInput::Address to directly provide the summa_contract's address.
     // For deployed contracts, if the address is stored in a config file,
     // you can alternatively use AddressInput::Path to specify the file's path.
```
enricobottazzi, line 32: Do we have an example of such config file? Can you add it to the comments if possible?

backend/src/tests.rs
```diff
@@ -180,62 +245,49 @@ mod test {
...
-            },
-            Asset {
-                asset_name: "USDT".to_string(),
-                chain: "ETH".to_string(),
-                amount: U256::from(556863),
-            },
-        ];
+        assert_eq!(liability_commitment_logs.len(), 0);
 
         // Send sovlecy proof to contract
```
enricobottazzi, line 263: fix comment

backend/examples/summa_solvency_flow.rs
```diff
@@ -47,30 +49,24 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-        anvil.chain_id(),
-        anvil.endpoint().as_str(),
-        AddressInput::Address(summa_contract.address()),
-        entry_csv,
-        asset_csv,
-        params_path,
-        1,
-    )
-    .unwrap();
+    let entries = get_sample_entries();
```
enricobottazzi, line 58: Can we add `entry.csv` file back here? I think there was a misunderstanding in one of my previous comments. I didn't mean to remove the usage of `entry.csv`. I just meants that this should no longer be an input for the `Round`  builder. But it's still ok to me to build the `mst` using `let merkle_tree =MerkleSumTree::new("src/merkle_sum_tree/csv/entry_16.csv")` and then pass it to `Round`

backend/src/sample_entries.rs
```diff
@@ -0,0 +1,89 @@
+use num_bigint::ToBigUint;
+use summa_solvency::merkle_sum_tree::Entry;
+
+pub fn get_sample_entries() -> Vec<Entry<2>> {
```
enricobottazzi, line 4: remove this file

backend/src/apis/round.rs
```diff
@@ -62,58 +58,56 @@ impl MstInclusionProof {
...
     [usize; N_ASSETS + 1]: Sized,
     [usize; 2 * (1 + N_ASSETS)]: Sized,
 {
-    pub fn new(
-        signer_key: &str,
-        chain_id: u64,
-        rpc_url: &str,
-        summa_address_input: AddressInput,
-        entry_csv_path: &str,
+    pub fn new<'a>(
```
enricobottazzi, line 76: The main point of the PR for me was that `Round` should accept as input an object binded to the `Tree` trait. Similarly to what you applied to `Solvency::init`. But I don't know why you applied it to `Solvency` which is now dismissed

backend/src/apis/round.rs
```diff
@@ -157,54 +149,26 @@ where
...
-            public_inputs: calldata.1,
+            trusted_setup: mst_inclusion_setup_artifacts,
         })
     }
 
     pub fn generate_proof_of_inclusion(
         &self,
         user_index: usize,
     ) -> Result<MstInclusionProof, &'static str> {
+        let merkle_proof = self.mst.generate_proof(user_index).unwrap();
```
enricobottazzi, line 163: As soon as you modify `Snapshot` to take an object with a Binded Trait `Tree` this object will no longer have the `get_entry` method as this method doesn't live on the `Tree` trait. I don't know what's the best way to handle this

contracts/src/Summa.sol
sifnoc: It's leftover from my audit comment, I will remove it.

backend/src/apis/round.rs
sifnoc: Ah, that's what you were looking for. Honestly, I was trying to figure out how to define the `Tree` trait as a concrete type in the `SolvencyCircuit` struct, but I realized it's impossible without using `Box`, which is kind of smart pointer.

enricobottazzi: Why is that?

Can't you define it like this? https://github.com/summa-dev/summa-solvency/blob/enrico-tree-trait/zk_prover/src/circuits/solvency.rs#L76

sifnoc: It because the `merkle_sum_tree` in the `solvency::init` evaluates some values, like `root_hash`, not assign into `SolvencyCircuit`. Anyway I fixed like this: 
```Rust
pub struct Snapshot<const LEVELS: usize, const N_ASSETS: usize, const N_BYTES: usize> {
    mst: Box<dyn Tree<N_ASSETS, N_BYTES>>, // Instead of MerkleSumTree<N_ASSETS, N_BYTES>,
    assets_state: [Asset; N_ASSETS],
    trusted_setup: SetupArtifacts,
}
```

backend/src/apis/round.rs
sifnoc: Well, as you already know I added `get_entry` method in `Tree` trait.
but for removing `get_entry` from `Tree` trait, we have to add `username` and `balances` fields on the `MerkleProof`.
the `entry` usually used to get `username` and `balances`  In the `MstInclusionCircuit`.
What do you think?

backend/src/tests.rs
sifnoc: Fixed at [5a9120f](https://github.com/summa-dev/summa-solvency/pull/179/commits/5a9120f25dfebf26826ba5799e4a53c571e8cf06)

backend/examples/summa_solvency_flow.rs
sifnoc: Updated at [5a9120f](https://github.com/summa-dev/summa-solvency/pull/179/commits/5a9120f25dfebf26826ba5799e4a53c571e8cf06)

backend/src/sample_entries.rs
sifnoc: Removed at [5a9120f](https://github.com/summa-dev/summa-solvency/pull/179/commits/5a9120f25dfebf26826ba5799e4a53c571e8cf06)

backend/src/apis/round.rs
sifnoc: Updated at [5a9120f](https://github.com/summa-dev/summa-solvency/pull/179/commits/5a9120f25dfebf26826ba5799e4a53c571e8cf06)

backend/examples/summa_solvency_flow.rs
sifnoc: Yes, there is "backend/src/contracts/deployments.json".
I will add comment

backend/src/apis/round.rs
enricobottazzi: Ok let's keep it as it is for now! 

If something comes up during the `AggregationMerkleSumTree` implementation, we can just come back to `summa-solvency` and fix that

contracts/src/Summa.sol
sifnoc: Removed [c575a29](https://github.com/summa-dev/summa-solvency/pull/179/commits/c575a299d5a585ea43204508a74541eda53af37e)

backend/examples/summa_solvency_flow.rs
sifnoc: Updated [c575a29](https://github.com/summa-dev/summa-solvency/pull/179/commits/c575a299d5a585ea43204508a74541eda53af37e)

backend/src/apis/round.rs
```diff
@@ -60,60 +56,58 @@ impl MstInclusionProof {
...
+            root_sums.push(U256::from_str_radix(&fp_str, 16).unwrap());
+        }
 
         self.signer
-            .submit_proof_of_solvency(
-                proof.public_inputs[0],
+            .submit_commitment(
+                mst_root,
+                root_sums,
                 self.snapshot.assets_state.to_vec(),
```
alxkzmn, line 110: Something to think about for the future: assets come from the same source as the user balances. E.g., when the CEX is building the tree, it knows that it's now querying the ETH balances or BTC balances. That's why I don't think that the assets would live in a separate CSV file. Imagine that when the CEX is querying the balances, it specifies the assets in the query. If we're sticking to CSV, I would imagine a CSV that looks like `userId,balance_ETH_ETH,balance_USDT_ETH,...`. Won't request changes now but Interesting to know what you think @enricobottazzi @sifnoc

alxkzmn: It's not really an asset "state" per se, it's kind of a mapping of what assets are submitted in the MST to leave the record inside the transaction.

alxkzmn (approved):

backend/src/apis/round.rs
enricobottazzi: I agree. It used to be a state when it also contained the asset_balances. Now it doesn't make sense for `asset.csv` to live in a separate file. Opening an issue.

Issue #166, opened 2023-10-31
Merkle Sum Tree Efficiency Improvement


enricobottazzi:
The current logic of hashing inside the Merkle Sum Tree is based on [DAPOL, paragraph 4.1](https://www.yji.me/publication/gpol/gpol.pdf). 

<img width="722" alt="Screenshot 2023-10-31 at 10 07 44" src="https://github.com/summa-dev/summa-solvency/assets/85900164/e96a035d-0699-4291-9cc4-4a95ec4000b1">

In particular, each middle node stores an hash and a balance. The hash is equal to `H(HLeftChild, BalanceLeftChild, HRightChild, BalanceRightChild)`, the balance is equal to `BalanceLeftChild + BalanceRightChild`

This is a fix to the original Maxwell tree in which the hash of a middle node is equal to `H(BalanceLeftChild + BalanceRightChild, HLeftChild, HRightChild)`. 

<img width="1264" alt="Screenshot 2023-10-31 at 10 10 29" src="https://github.com/summa-dev/summa-solvency/assets/85900164/8d968065-c5c3-4c19-870c-b8e54f2b7375">

This Maxwell tree has a vulnerability described in [Broken Proofs of Solvency in Blockchain Custodial Wallets and Exchanges, paragraph 4.1.](https://www.yji.me/publication/pol-attacks/pol-attacks.pdf). The exchange could set the balance of a middle node as `Max(BalanceLeftChild, BalanceRightChild)` therefore reducing its total liabilities.

When generating a Merkle Proof for a user, the Exchange could modify the balances of the sibling nodes on the fly such that the sum matches `Max(BalanceLeftChild, BalanceRightChild)`.  

<img width="861" alt="Screenshot 2023-10-31 at 10 16 41" src="https://github.com/summa-dev/summa-solvency/assets/85900164/8f404cec-362f-4af4-8712-6a751d89428b">

This issue is possible because Alice `h2`. If she could open `h2=h(10, Bob, idBob)`, she could tell that the `v2` should be 10 and not 5. Unfortunately, in such architecture where each user is receiving a Merkle Proof, we don't want Alice to be able to open Bob's hash because this would reveal the identity and the balance of another user.

[...]

PR #188, opened 2023-11-21
`Mst` efficiency improvement

enricobottazzi:
#166 

Major breaking changes in this PR: 

- Logic of building the middle nodes of the MerkleSumTree is modified. Before the pre-image of the hash was equal to `balance_left, hash_left, balance_right, hash_right` now it is `balance_left + balance_right, hash_left, hash_right`
- Further constraints added to the circuit. Before the circuit would just take `sibling_sums` and `sibling_hashes` to be used inside the witness. Now we need to open the hash of each sibling. Because of that the `MerkleProof` now has fields`sibling_leaf_node_hash_preimage` and `sibling_middle_node_hash_preimages` rather than `sibling_hashes` and `sibling_sums`. The logic of the circuit is modified to build each `sibling_hash` inside the circuit and constraint that the `balances` associated to a sibling are the same used as input to build the hash

To do:

- As I modified the `Tree` trait, it might also have an impact on `AggregationMerkleSumTree`

Benchmarks:

- Reduced Merkle Sum Tree generation time (2^20 users, 1 asset) -> 73.695 s  vs 88.182s
- Increased Proof of Inclusion generation time -> 517.98 ms vs 465.61 ms

Given that the focus of V1 is fast commitment time. I think that this is a reasonable tradeoff.


zk_prover/src/merkle_sum_tree/tree.rs
```diff
@@ -64,14 +142,49 @@ pub trait Tree<const N_ASSETS: usize, const N_BYTES: usize> {
...
+        [usize; N_ASSETS + 2]: Sized,
     {
-        let mut node = proof.leaf.clone();
+        let mut node = proof.entry.compute_leaf();
+
+        let sibling_leaf_node_balances = proof.sibling_leaf_node_hash_preimage[1..]
+            .try_into()
+            .unwrap();
+
+        let sibling_leaf_node_hash = poseidon_entry::<N_ASSETS>(
```
alxkzmn, line 153: poseidon_entry as well as poseidon_node are specific to Node creation and are always used in that context, so their implementations can be placed directly in the Node (no need to have them inside a separate file in utils). For example, here you first create the hash of sibling_leaf_node and then create the sibling_leaf_node Node using this hash, not using the hash for anything else. Meanwhile, Node already has the ::leaf constructor that takes username and balances as args. Moving both hashing operations into the Node would remove confusion over how to create the node.

zk_prover/src/merkle_sum_tree/mod.rs
```diff
@@ -7,11 +7,15 @@ pub mod utils;
 use halo2_proofs::halo2curves::bn256::Fr as Fp;
 
 #[derive(Clone, Debug)]
-pub struct MerkleProof<const N_ASSETS: usize, const N_BYTES: usize> {
```
sifnoc, line 10: To clarify the meaning of the numbers adjacent to `N_ASSETS`, I would like to suggest an update as follows:

```Rust
#[derive(Clone, Debug)]
pub struct MerkleProof<const N_ASSETS: usize, const N_BYTES: usize>
where
    [usize; 1 + N_ASSETS]: Sized, // Adding 1 for the username
    [usize; N_ASSETS + 2]: Sized, // Adding 2 for two hashes
{
    pub entry: Entry<N_ASSETS>,
    pub root: Node<N_ASSETS>,
    pub sibling_leaf_node_hash_preimage: [Fp; 1 + N_ASSETS],
    pub sibling_middle_node_hash_preimages: Vec<[Fp; N_ASSETS + 2]>,
    pub path_indices: Vec<Fp>,
}
```

This suggestion makes the code more intuitive, and it should help newcomers more easily understand the hash inputs.

backend/src/apis/round.rs
```diff
@@ -120,7 +126,10 @@ where
         mst: Box<dyn Tree<N_ASSETS, N_BYTES>>,
         asset_csv_path: &str,
         params_path: &str,
-    ) -> Result<Snapshot<LEVELS, N_ASSETS, N_BYTES>, Box<dyn std::error::Error>> {
+    ) -> Result<Snapshot<LEVELS, N_ASSETS, N_BYTES>, Box<dyn std::error::Error>>
+    where
+        [(); N_ASSETS + 2]: Sized,
```
sifnoc, line 131: Would you be able to update the `where` clause just under `Snapshot` too?

backend/src/apis/round.rs
```diff
@@ -63,12 +63,15 @@ where
         asset_csv_path: &str,
         params_path: &str,
         timestamp: u64,
-    ) -> Result<Round<'a, LEVELS, N_ASSETS, N_BYTES>, Box<dyn Error>> {
+    ) -> Result<Round<'a, LEVELS, N_ASSETS, N_BYTES>, Box<dyn Error>>
```
sifnoc, line 66: Would you be able to update the `where` clause just under `Round` too?

zk_prover/src/circuits/merkle_sum_tree.rs
```diff
@@ -107,23 +115,26 @@ where
 /// * `poseidon_middle_config`: Configuration for the poseidon hash function with WIDTH = 2 and RATE = 1 and input length of 2 * (1 + N_ASSETS). Needed to perform hashings from the leaf to the root.
```
sifnoc, line 115: The comment should likely be updated to `N_ASSETS + 2` instead of `2 * (1 + N_ASSETS)`

sifnoc (changes requested):
This is a good improvement, especially for large entry sets.
Also, thank you for the minor updates in the backend.

zk_prover/src/merkle_sum_tree/tree.rs
enricobottazzi: Thanks for the suggestion! Changes applied in #fc44e01. I hope that now it looks cleaner

zk_prover/src/merkle_sum_tree/mod.rs
enricobottazzi: Applied in [94404ed](https://github.com/summa-dev/summa-solvency/pull/188/commits/94404ed931227f9fa51e6a4be17ffe7ee4e564ae)

zk_prover/src/circuits/merkle_sum_tree.rs
enricobottazzi: Good catch! Thanks for that

enricobottazzi:
I updated the code base according to your comments. A further thing that I noticed while reviewing the codebase is that there was an API `generate_leaf_hash` in `zk_prover` that was only used inside `backend/examples`. I therefore decided to move it back to `backend` as part of our effort to reduce the number of `utils` API inside `zk_prover`.

You might notice that the tests are failing in the Github Action. This is the same issue that @alxkzmn encountered and solved in #192. Therefore I propose to merge that PR first.

sifnoc (approved):

zk_prover/src/merkle_sum_tree/node.rs
```diff
@@ -1,34 +1,55 @@
...
 #[derive(Clone, Debug)]
 pub struct Node<const N_ASSETS: usize> {
     pub hash: Fp,
     pub balances: [Fp; N_ASSETS],
 }
 impl<const N_ASSETS: usize> Node<N_ASSETS> {
+    /// Builds a leaf-level node of the MST
+    /// The leaf node hash is equal to `H(username, balance[0], balance[1], ... balance[N_ASSETS - 1])`
+    /// The balances are equal to `balance[0], balance[1], ... balance[N_ASSETS - 1]`
+    pub fn leaf(username: &BigUint, balances: &[BigUint; N_ASSETS]) -> Node<N_ASSETS>
```
alxkzmn, line 16: In the true spirit of DRY this could just call the `Node::leaf_node_from_preimage()` after converting the args :) Same with the middle calling Node::middle_from_preimage inside itself. Won't request changes, we can do this minor enhancement later.

enricobottazzi: Let's DRY! 

Gonna apply the changes now (:

zk_prover/src/merkle_sum_tree/utils/build_tree.rs
```diff
@@ -70,7 +70,18 @@ where
...
+            let mut hash_preimage = [Fp::zero(); N_ASSETS + 2];
+
+            for (i, balance) in hash_preimage.iter_mut().enumerate().take(N_ASSETS) {
+                *balance =
+                    tree[level - 1][index].balances[i] + tree[level - 1][index + 1].balances[i];
+            }
+
+            hash_preimage[N_ASSETS] = tree[level - 1][index].hash;
+            hash_preimage[N_ASSETS + 1] = tree[level - 1][index + 1].hash;
+            Node::middle_node_from_preimage(&hash_preimage)
```
alxkzmn, line 83: Actually that's why we could have both constructors. I just meant that one could call the other inside itself.

enricobottazzi: Oh I see, I misunderstood your point

enricobottazzi: Applied changes in last commit!

PR #191, opened 2023-11-24
Add circuit audit comments

alxkzmn:
There are two circuit audit comments starting with // Audit:

There are also minor typo fixes as well as some changes suggested by clippy.

### DIFF
--- a/zk_prover/src/chips/merkle_sum_tree.rs
+++ b/zk_prover/src/chips/merkle_sum_tree.rs
@@ -20,7 +20,7 @@ pub struct MerkleSumTreeConfig {
 /// Chip that performs various constraints related to a Merkle Sum Tree data structure such as:
 ///
 /// * `s * swap_bit * (1 - swap_bit) = 0` (if `bool_and_swap_selector` is toggled). It basically enforces that swap_bit is either a 0 or 1.
-/// * `s * (swap_bit * 2 * (elelment_r_cur - elelment_l_cur)  - (elelment_l_next - elelment_l_cur) - (elelment_r_cur - elelment_r_next)) = 0`. Enforces that if the swap_bit is equal to 1, the values will be swapped on the next row (if `bool_and_swap_selector` is toggled).
+/// * `s * (swap_bit * 2 * (element_r_cur - element_l_cur)  - (element_l_next - element_l_cur) - (element_r_cur - element_r_next)) = 0`. Enforces that if the swap_bit is equal to 1, the values will be swapped on the next row (if `bool_and_swap_selector` is toggled).
 /// If the swap_bit is equal to 0, the values will remain the same on the next row (if `bool_and_swap_selector` is toggled).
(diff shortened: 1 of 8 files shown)


zk_prover/src/chips/merkle_sum_tree.rs
```diff
@@ -59,6 +59,22 @@ impl<const N_ASSETS: usize> MerkleSumTreeChip<N_ASSETS> {
...
+            // that satisfy the composite equation without satisfying the intended individual constraints.
+            // For the case where swap_bit = 0:
+            // element_l_cur − element_l_next + element_r_next − element_r_cur = 0
+            // Here, it's theoretically possible to cheat by picking element_l_cur, element_l_next,
+            // element_r_cur, and element_r_next such that they don't individually satisfy
+            // element_l_cur − element_l_next = 0 and element_l_cur − element_l_next = 0
+            // but still satisfy the combined equation.
+            // For example, if element_l_cur = 3 and element_l_next = 1, and element_r_cur = 2 and element_r_next = 4,
+            // the composite equation would still be satisfied because:
+            // 3 − 1 + 4 − 2 = 0
```
sifnoc, line 75: I understand the issue you've pointed out. A malicious prover could potentially exploit the switch in values at this point:
https://github.com/summa-dev/summa-solvency/blob/7489c53ebdd4dd8bbcb3b45490cccee6426fc7c5/zk_prover/src/chips/merkle_sum_tree.rs#L223-L235

However, considering the example provided, it seems that the constraint would not be satisfied.

$$ 1 * 2 * ( 3 - 2 ) - ( 1 - 3) - ( 2 - 4) = 2 \neq 0 $$

Please correct me if I'm missing something or if I'm wrong.

sifnoc (changes requested):

zk_prover/src/chips/merkle_sum_tree.rs
alxkzmn: The example is for the case when the swap bit is zero: swap_bit * 2 * (element_r_cur - element_l_cur) - (element_l_next - element_l_cur) - (element_r_cur - element_r_next) = - (element_l_next - element_l_cur) - (element_r_cur - element_r_next) = element_l_cur - element_l_next + element_r_next - element_r_cur

alxkzmn: For example, 3 - 4 + 2 - 1 = 0.

sifnoc: I see. For better understanding in the example, perhaps assigning 5 to `element_l_next` instead of 1 would make it clearer.

enricobottazzi (commented):
I replied to the two audit issues. Let's keep the PR open until we reach an agreement especially for the swapping constraint issue. When we are good, I'm gonna apply the solution for the two issues (together with the clippy fixes) in a separated PR

zk_prover/src/chips/merkle_sum_tree.rs
alxkzmn: Yes, I think this is the solution.

sifnoc: Agreed

Issue #197, opened 2023-11-28
Fix `csv` files folder for Benchmark


enricobottazzi:
Benchmarks are broken now. 

The csv files zip should be fixed according to the new expected csv format as a consequence of #192

Furthermore the naming of the files should adapt to the new convention #200. So replace `asset` with `currency`

Issue #167, opened 2023-10-31
ZK Proof of Inclusion Efficiency improvement


enricobottazzi:
The [ZK Proof of Inclusion circuit](https://github.com/summa-dev/summa-solvency/blob/master/zk_prover/src/circuits/merkle_sum_tree.rs) constrains the balances to be in the range defined by `N_BYTES`. This is necessary to avoid overflow during the summation.

<img width="492" alt="Screenshot 2023-10-31 at 11 16 13" src="https://github.com/summa-dev/summa-solvency/assets/85900164/d0baa201-08b3-4d5c-a8b2-269fe0bf4bc2">

For example, considering the CEX generating a ZK Proof of Inclusion for Carl, the circuit would check for each hashing level that [the balances are in the correct range](https://github.com/summa-dev/summa-solvency/blob/master/zk_prover/src/circuits/merkle_sum_tree.rs#L304-L337).

Eventually, the circuit also performs [the range check on the root balance](https://github.com/summa-dev/summa-solvency/blob/master/zk_prover/src/circuits/merkle_sum_tree.rs#L377-L380)

In particular, it would perform the range check on `10 ETH, 15 ETH, 70 ETH, 25 ETH, 95 ETH, 140 ETH, 235 ETH`.  This accounts for `2*(N_LEVELS - 1) + 1` range checks to be performed inside the circuit. 

What we can do instead is just to perform the range check on the `entry.balance` and on the `path_element_balances`, which in that case would be `10 ETH, 15 ETH, 70 ETH, 140 ETH`. All the other balances are computed as the sum of these, therefore, setting an appropriate `N_BYTES` would make an overflow of the circuit prime impossible. 

In the example, if the range check is set to `N_BYTES = 8`. The max value that `entry.balance` and `path_element_balances` can have is 2^64 - 1.  

- After the first summation (between the leaf nodes), the max value of the balance in the computed node is 2*(2^64 - 1).
- After the second summation (between the computed node and the sibling node), the max value of the balance in the newly computed node is 2^64 - 1 + 2*(2^64 - 1)
- After the third summation (between the computed node and the sibling node), the max value of the balance in the root node is 2^64 -1 + 2^64 - 1 + 2*(2^64 - 1). 

This is the python algorithm for calculating `maxRootBalance`

```
def calculate_max_root_balance(n_bytes, n_levels):
    # The max value that can be stored in a leaf node or a sibling node, according to the constraint set in the circuit
[...]

PR #198, opened 2023-11-28
Circuit improvements

enricobottazzi:
- Apply audit fixes #191 
- Apply improvement on range checks #167 

On the last part there's still two missing things:

- Enforcing that no overflow can happen inside the circuit, based on the chosen parameters `N_BYTES` and `LEVELS` on the smart contract level. (this is better described in the issue) cc: @alxkzmn 
- Update the benchmarks. For now the benchmarks are broken due to the new csv format. I opened an issue for that (#197). I think we can still go on and merge this PR and update the benchmarks later @sifnoc


enricobottazzi:
- @alxkzmn to add possibility of recording the generic parameters of the `[MstInclusionCircuit](https://github.com/summa-dev/summa-solvency/blob/master/zk_prover/src/circuits/merkle_sum_tree.rs#L29)` on the Summa contract. 
- @enricobottazzi to create a script to verify that the combination circuit + generic parameters + trusted setup leads to the same smart contract

enricobottazzi:
@alxkzmn thanks for sorting this out! I think that the value that you hardcoded for the merkle tree levels is not correct.  

These are the parameters that we are using for building the verifier (https://github.com/summa-dev/summa-solvency/blob/circuit-improvement-enrico/zk_prover/examples/gen_inclusion_verifier.rs).

Can you update them, adding `N_ASSETS` as well and add in the comments a link to the script that we use to generate such verifier contract

zk_prover/src/chips/merkle_sum_tree.rs
```diff
@@ -20,7 +20,7 @@ pub struct MerkleSumTreeConfig {
 /// Chip that performs various constraints related to a Merkle Sum Tree data structure such as:
 ///
 /// * `s * swap_bit * (1 - swap_bit) = 0` (if `bool_and_swap_selector` is toggled). It basically enforces that swap_bit is either a 0 or 1.
-/// * `s * (swap_bit * 2 * (elelment_r_cur - elelment_l_cur)  - (elelment_l_next - elelment_l_cur) - (elelment_r_cur - elelment_r_next)) = 0`. Enforces that if the swap_bit is equal to 1, the values will be swapped on the next row (if `bool_and_swap_selector` is toggled).
+/// * `s * (swap_bit * 2 * (element_r_cur - element_l_cur)  - (element_l_next - element_l_cur) - (element_r_cur - element_r_next)) = 0`. Enforces that if the swap_bit is equal to 1, the values will be swapped on the next row (if `bool_and_swap_selector` is toggled).
```
alxkzmn, line 23: The description is outdated

enricobottazzi: Fixed!

sifnoc (approved):
Looks good to me! but I left small thing what i found.

enricobottazzi:
Thanks @sifnoc, applied changes

zk_prover/src/circuits/traits.rs
```diff
@@ -30,4 +30,24 @@ pub trait CircuitBase {
             |mut region| region.assign_advice(|| "value", advice_col, 0, || Value::known(value)),
         )
     }
+
+    /// Loads the lookup table with values from `0` to `2^8 - 1`
+    fn load(&self, layouter: &mut impl Layouter<Fp>, column: Column<Fixed>) -> Result<(), Error> {
```
alxkzmn, line 35: Technically this is not a problem, but this method is very specific to the implementation of our MST. It makes `CircuitBase` not a common trait anymore but a specific API for this case, so either naming and description don't correspond to the function anymore, or the method should be inside the MST circuit implementation.

alxkzmn: I guess the trait is a leftover from the time when we had two circuits

PR #199, opened 2023-11-29
Implement a self-hosted runner dedicated to the Rust workflow

ntampakas:
**Info**
With this PR, a new self-hosted runner is introduced for the Rust workflow, which is located in AWS.

**Changes**
- rust workflow was modified to run on the self-hosted runner. A new job was created as well, that wakes up the runner and uses GH' s OIDC provider to get short-lived AWS credentials required for the specific actions.
- wakeup.sh script was created, to launch the specific runner.

The runner in question, is a C6a class instance with 48vCPUs and 96GB of RAM.
Total execution time was decreased from ~15' -> 4'.

### DIFF
--- a/.github/scripts/wakeup.sh
+++ b/.github/scripts/wakeup.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -e
+
+runner="i-0e04845bff4576909"
+
+while true; do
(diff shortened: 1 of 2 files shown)


enricobottazzi (approved):
This is definitely faster. 🚀 LGTM

sifnoc (approved):
Finally, we get a boost! Thank

alxkzmn (approved):

PR #206, opened 2023-12-04
Modified new method in MerkleSumTree in mst

sifnoc:
The current `new` method in `MerkleSumTree` requires a `csv_path` parameter. However, for the Summa Aggregation project, there's a need for an actual `new` method that aligns with the implementation shown here: [Summa Aggregation - json_mst](https://github.com/summa-dev/summa-aggregation/blob/4a4d71b619251382feb2e85b7992835221ecc2d6/src/json_mst.rs#L116-L121).

```Rust
MerkleSumTree::<N_ASSETS, N_BYTES> {
    root,
    nodes,
    depth: self.depth,
    entries,
    is_sorted: self.is_sorted,
}
```

### DIFF
--- a/backend/examples/summa_solvency_flow.rs
+++ b/backend/examples/summa_solvency_flow.rs
@@ -59,7 +59,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Initialize the `Round` instance to submit the liability commitment.
     let params_path = "ptau/hermez-raw-11";
     let entry_csv = "../csv/entry_16.csv";
-    let mst = MerkleSumTree::new(entry_csv).unwrap();
+    let mst = MerkleSumTree::from_csv(entry_csv).unwrap();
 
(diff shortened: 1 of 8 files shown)


sifnoc:
> @sifnoc I already mentioned it in the latest review of `summa-aggregation` but can't you just use the `mst.from_entries` method ?

Thank you for pointing this out again in here(It must be hard to find in PR). I apologize for any confusion caused earlier. I addressed this point in our previous discussion here: https://github.com/summa-dev/summa-aggregation/pull/3#discussion_r1414206368.

To clarify, we are indeed using `MerkleSumTree::from_entries` in the summa-aggregation project, as you can see in the code here: https://github.com/summa-dev/summa-aggregation/blob/6b1a14550352508fa800747c8318ce99dbe7cfd5/src/mini_tree_generator.rs#L35

enricobottazzi (approved):
LGTM! 

Can you just rename the `new` method with something like `from_params` ?

alxkzmn (approved):

Issue #216, opened 2023-12-07
Add Zero Element to MST


enricobottazzi:
- [ ] Code
- [ ] Docs

PR #217, opened 2023-12-07
feat: add zero element to the tree

enricobottazzi:
#216


sifnoc (approved):
That's good!
By the way, a similar modification is needed for AggregationMerkleSumTree in summa-aggregation.
We can also avoid calculating for 0 elements by providing a pre-computed hash result in the futures.

enricobottazzi:
> We can also avoid calculating for 0 elements by providing a pre-computed hash result in the futures.

Right, I think this is a good idea. Let me try to add it to the current PR.

enricobottazzi:
@sifnoc Updated in 8311509. As I was working on that, I realized that the precomputed leaf hash is different for any `N_CURRENCIES` value. Therefore, I just added the support for the precomputed hash in the case `N_CURRENCIES = 1`. For `N_CURRENCIES` greater than 1, there's no precomputed hash

sifnoc (approved):

zk_prover/src/merkle_sum_tree/mst.rs
```diff
@@ -133,6 +113,14 @@ impl<const N_CURRENCIES: usize, const N_BYTES: usize> MerkleSumTree<N_CURRENCIES
 
         let mut nodes = vec![];
 
+        // Pad the entries with empty entries to make the number of entries equal to 2^depth
+        if entries.len() < 2usize.pow(depth as u32) {
```
alxkzmn, line 117: I wonder if instead of stuffing the rest of the tree we could only add the zero node when the right hasher input is missing on any level. This way there would be less computation over the empty branches.

alxkzmn (approved):

zk_prover/src/merkle_sum_tree/mst.rs
enricobottazzi: Thanks for the idea, opened an issue 4 that!

PR #218, opened 2023-12-08
Remove unused return variables

alxkzmn:
The variables are actually not used after the assignment.


enricobottazzi (approved):
Thanks for the clean up 🧹

sifnoc (approved):
Great catch!

PR #224, opened 2023-12-11
Create KZG multi-openings & verify user ID at the specified index

### LINKED ISSUE #210: KZG multiopening

alxkzmn:
- The difference with the single-opening proof is the query batching - multiple polynomials are queried together, which in our case allows to batch the user or grand sum queries to multiple cryptocurrency balance polynomials.
- The inclusion of a user ID at the correct user index is now also verified (note that the user ID advice column can stay blinded because we don't need the special grand sum property of the polynomial). 

Closes #210 & #222.


kzg_prover/src/circuits/tests.rs
```diff
@@ -85,26 +85,26 @@ mod test {
...
         );
 
-        // The Custodian creates a KZG proof of the 4th user balances inclusion
+        // The Custodian creates a KZG multi-proof of the 4th user ID & balances inclusion
         let user_index = 3_u16;
 
-        let balance_column_range = 1..N_CURRENCIES + 1;
-        let balance_opening_proofs = open_user_balances::<N_CURRENCIES>(
+        let column_range = 0..N_CURRENCIES + 1;
+        let openings_multi_proof = open_user_points::<N_CURRENCIES>(
```
enricobottazzi, line 103: Again, what's inside this multi-proof?

kzg_prover/src/circuits/utils.rs
```diff
@@ -192,82 +185,74 @@ pub fn verify_grand_sum_openings<const N_CURRENCIES: usize>(
...
 pub fn verify_user_inclusion<const N_CURRENCIES: usize>(
     params: &ParamsKZG<Bn256>,
     zk_snark_proof: &[u8],
-    balance_opening_proofs: Vec<Vec<u8>>,
+    balance_opening_multi_proof: &[u8],
     balance_column_range: Range<usize>,
     omega: Fp,
     user_index: u16,
-) -> (Vec<bool>, Vec<BigUint>) {
+) -> (bool, Vec<BigUint>) {
```
enricobottazzi, line 218: Maybe it's more intuitive to separate userID and balances when returned from `verify_user_inclusion`

kzg_prover/src/circuits/tests.rs
```diff
@@ -122,39 +122,64 @@ mod test {
...
+                assert_eq!(
+                    *fourth_user_csv_entry.balances().get(i - 1).unwrap(),
+                    id_and_balance_values[i]
+                );
+            }
         }
+
+        let balance_column_range = 1..N_CURRENCIES + 1;
+
+        // Test failure case with the wrong group generator
```
enricobottazzi, line 170: I suggest to separate the test failure in a different test vector. Also probably it makes sense to add more failure test and assert their specific unexpected behaviour.

kzg_prover/src/circuits/tests.rs
```diff
@@ -85,26 +85,26 @@ mod test {
             full_prover(&params, &pk, circuit.clone(), vec![vec![]]);
```
enricobottazzi, line 85: Specify what is this ZK proof and how it differs from the other proofs that we are generating later

kzg_prover/src/circuits/tests.rs
```diff
@@ -122,39 +122,64 @@ mod test {
         // Both the Custodian and the Verifier know what column range are the balance columns
```
enricobottazzi, line 122: Is `poly_degree` the same as number of users?

kzg_prover/src/circuits/utils.rs
```diff
@@ -313,26 +307,27 @@ pub fn verify_opening<
     E: EncodedChallenge<Scheme::Curve>,
     T: TranscriptReadBuffer<&'a [u8], Scheme::Curve, E>,
     Strategy: VerificationStrategy<'params, Scheme, V, Output = Strategy>,
+    const N_POINTS: usize,
```
enricobottazzi, line 310: Does `verify_opening` need to be public?

enricobottazzi (changes requested):
/

kzg_prover/src/circuits/utils.rs
```diff
@@ -192,82 +185,74 @@ pub fn verify_grand_sum_openings<const N_CURRENCIES: usize>(
...
-        }
-    }
-    (verification_results, constant_terms)
+    let (verified, constant_terms) = verify_opening::<
+        KZGCommitmentScheme<Bn256>,
+        VerifierSHPLONK<'_, Bn256>,
+        Challenge255<G1Affine>,
+        Blake2bRead<_, _, Challenge255<_>>,
+        AccumulatorStrategy<_>,
+        N_CURRENCIES,
```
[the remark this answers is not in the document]
alxkzmn: Yes, we're opening the commitments to multiple polynomials at the same time (they're written into a single transcript so it may be not that obvious).

kzg_prover/src/circuits/tests.rs
```diff
@@ -85,26 +85,26 @@ mod test {
...
 
         // Both the Custodian and the Verifier know what column range are the balance columns
+        // (The first column is the user IDs)
         let balance_column_range = 1..N_CURRENCIES + 1;
 
-        // The Custodian makes an opening at x = 0 for the Verifier
-        // (The first column is the user IDs)
-        let kzg_proofs = open_grand_sums::<N_CURRENCIES>(
+        // The Custodian makes multi-openings at x = 0 for the Verifier
+        let grand_sums_multi_proof = open_grand_sums::<N_CURRENCIES>(
```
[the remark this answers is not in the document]
alxkzmn: Just KZG commitments to multiple polynomials.

kzg_prover/src/circuits/utils.rs
alxkzmn: Not sure about that because all the KZG commitments are treated in the same way here. The caller knows what columns are balances and what is the user ID, and the caller could pass these column indices into the function, but the only thing the function would do with them is using them to split the output for the caller (who can do it themselves because of knowing what columns they are).

kzg_prover/src/circuits/tests.rs
alxkzmn: No, it can be different because hypothetically there can be something else in the circuit, or simply the number of users is not a power of two.

kzg_prover/src/circuits/utils.rs
alxkzmn: No, will remove the modifier.

kzg_prover/src/circuits/tests.rs
alxkzmn: Will move it into a separate negative test and create an issue for the additional negative tests.

alxkzmn: https://github.com/summa-dev/summa-solvency/issues/229

kzg_prover/src/circuits/tests.rs
alxkzmn: Actually they are not the same right now because K = 9 so there are 2^9 points that the advice polynomials interpolate, but only 16 of them are filled with the user balances during the test.

kzg_prover/src/circuits/utils.rs
enricobottazzi: I understand your explanation and makes sense to me!

kzg_prover/src/circuits/tests.rs
alxkzmn: I will document all the new KZG functions in `utils` not to explain in details how the KZG proofs work in the test.

kzg_prover/src/circuits/tests.rs
```diff
@@ -122,44 +123,138 @@ mod test {
...
+        assert!(verified);
         for i in 0..N_CURRENCIES {
-            assert!(verified[i]);
             assert_eq!(csv_total[i], grand_sum[i]);
         }
 
-        let balance_column_range = 1..N_CURRENCIES + 1;
-        let (balances_verified, balance_values) = verify_user_inclusion::<N_CURRENCIES>(
+        let column_range = 0..N_CURRENCIES + 1;
+        // The Verifier verifies the inclusion of the 4th user balances
```
enricobottazzi, line 142: I would say 4th user entry, which is made of used ID and balances

enricobottazzi (approved):

kzg_prover/src/circuits/utils.rs
```diff
@@ -192,82 +227,104 @@ pub fn verify_grand_sum_openings<const N_CURRENCIES: usize>(
...
+/// # Returns
+/// * `bool` - whether the user entry openings are verified correctly
+/// * `Vec<BigUint>` - the evaluations of the advice polynomials at the point corresponding to the user index
+pub fn verify_user_inclusion<const N_POINTS: usize>(
     params: &ParamsKZG<Bn256>,
     zk_snark_proof: &[u8],
-    balance_opening_proofs: Vec<Vec<u8>>,
-    balance_column_range: Range<usize>,
+    balance_opening_batch_proof: &[u8],
+    column_range: Range<usize>,
```
sifnoc, line 272: Could we use the range derived from `user_index` and `N_POINTS` instead of using `column_range`?

sifnoc (changes requested):
Great! Updated comments would help with understanding. 
However, I've also left some additional minor requests and favors in the comments.

kzg_prover/src/circuits/utils.rs
alxkzmn: No. User index is an index "along" the polynomial, which means a row index of advice columns. N_POINTS are the number of KZG commitments we're dealing with, but we cannot be sure where it starts. For example, if later we allocate the initial advice columns for something other that user ID, that would move the advice columns of interest to the right. I tried to make the function API general enough to be independent of our circuit implementation (because it also lives in a different module and is circuit-agnostic).

sifnoc: I understood, thanks for explanation.

sifnoc (approved):

PR #233, opened 2023-12-15
Remove `N_BYTES` const generic from `UnivariateGrandSum`

enricobottazzi:
#227


kzg_prover/src/chips/range/range_check.rs
```diff
@@ -1,113 +1,112 @@
...
 /// `z` is the advice column that contains the element to be checked.
 ///
-/// `z = 0x1f2f3f4f`
-/// `zs[0] = (0x1f2f3f4f - 0x4f) / 2^8 = 0x1f2f3f`
-/// `zs[1] = (0x1f2f3f - 0x3f) / 2^8 = 0x1f2f`
-/// `zs[2] = (0x1f2f - 0x2f) / 2^8 = 0x1f`
-/// `zs[3] = (0x1f - 0x1f) / 2^8 = 0x00`
+/// `z = 0x1f2f3f4f5f6f7f8f`
+/// `zs[0] = (0x1f2f3f4f5f6f7f8f - 0x7f8f) / 2^16 = 0x1f2f3f4f5f6`
+/// `zs[1] = (0x1f2f3f4f5f6 - 0xf5f6) / 2^16 = 0x1f2f3f4`
```
sifnoc, line 33: `zs[0]`and `zs[1]` are missing `f` , It should be ends with `f` like:
= `0x1f2f3f4f5f6`
= `0x1f2f3f4f`

sifnoc (changes requested):
I am a bit surprised that this implementation of the range-check chip is approximately 30% faster than the previous 8-byte version, even under the same K=17 condition.

I found tiny ones in comments. otherwise, everything looks good to me

sifnoc (approved):

alxkzmn (approved):

PR #246, opened 2023-12-28
Minor fixes to V2

enricobottazzi:
The main update is in the variable name `poly_degree` that becomes `poly_length` according to the discussion here -> https://hackmd.io/@summa/BkglBWsDp?comment=fa174e40-82cb-464d-bba4-fd9c6dfa8019&utm_source=comment-card&utm_medium=icon

On top of that I discover a more precise cargo clippy version (https://www.reddit.com/r/learnrust/comments/13pvxzu/comment/jlbk9mg/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) so I applied some of the changes suggested. I think this is useful to suggest when variable can be passed as reference to a function because they are not actually consumed within the function body.


alxkzmn (approved):

sifnoc (approved):
I hope for improved benchmark results with this update, especially with the application of references instead of clones :)

PR #249, opened 2024-01-02
Avoid calculating known polynomial evaluations

alxkzmn:
The refactoring haven't led to performance increase but I think that it's still better not to perform unnecessary computations.

### DIFF
--- a/kzg_prover/benches/kzg.rs
+++ b/kzg_prover/benches/kzg.rs
@@ -60,7 +60,7 @@ fn bench_kzg<const K: u32, const N_USERS: usize, const N_CURRENCIES: usize, cons
 
     let (zk_snark_proof, advice_polys, omega) = full_prover(&params, &pk, circuit, &[vec![]]);
 
-    let poly_length = u64::try_from(advice_polys.advice_polys[0].len()).unwrap();
+    let poly_length = 2 ^ u64::from(K);
 
     c.bench_function(&opening_grand_sum_bench_name, |b| {
         b.iter_batched(
--- a/kzg_prover/benches/kzg.rs
+++ b/kzg_prover/benches/kzg.rs
@@ -66,7 +66,7 @@ fn bench_kzg<const K: u32, const N_USERS: usize, const N_CURRENCIES: usize, cons
         b.iter_batched(
             || 1..N_CURRENCIES + 1,
             |balance_column_range| {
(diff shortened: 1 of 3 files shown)


kzg_prover/src/circuits/utils.rs
```diff
@@ -141,6 +141,7 @@ pub fn open_grand_sums<const N_CURRENCIES: usize>(
     advice_blinds: &[Blind<Fp>],
     params: &ParamsKZG<Bn256>,
     balance_column_range: Range<usize>,
+    polynomial_evaluations: &[Fp],
```
enricobottazzi, line 144: Can you add the description of this new input in the comments? For all the apis that are now using `polynomial_evaluations` as input

enricobottazzi (changes requested):

kzg_prover/benches/kzg.rs
```diff
@@ -59,6 +72,11 @@ fn bench_kzg<const K: u32, const N_USERS: usize, const N_CURRENCIES: usize, cons
                     &advice_polys.advice_blinds,
                     &params,
                     balance_column_range,
+                    csv_total
```
sifnoc, line 75: I just wonder that you want to include the polynomial evaluation inside this bench, right? if not, you can offload this computation on `setup` input instead of `routine`.

alxkzmn: As we discussed, it's probably negligible.

kzg_prover/src/circuits/tests.rs
```diff
@@ -59,25 +60,41 @@ mod test {
         // (The first column is the user IDs)
         let balance_column_range = 1..N_CURRENCIES + 1;
 
+        // The Custodian communicates the polynomial length to the Verifier
+        let poly_length = u64::try_from(advice_polys.advice_polys[0].len()).unwrap();
```
enricobottazzi, line 64: Can we retrieve `poly_length` from 2^K instead of this? There shouldn't be any substantial difference but using K reflects better what we are gonna do in production, namely fetching the K from the verifier smart contract.

(This also applies to `kzg.rs` in `benches`

enricobottazzi (approved):

kzg_prover/src/circuits/tests.rs
alxkzmn: Done

sifnoc (approved):

PR #251, opened 2024-01-03
Change username column to unblinded advice

alxkzmn:
As we observed in the benchmarks, there was a discrepancy between the user opening time and the grand sum opening time. I found out that the discrepancy was due to the user ID column being the regular advice column. After making it an unblinded advice column the discrepancy disappears.

### DIFF
--- a/kzg_prover/src/circuits/univariate_grand_sum.rs
+++ b/kzg_prover/src/circuits/univariate_grand_sum.rs
@@ -54,7 +54,7 @@ where
     [(); N_CURRENCIES + 1]:,
 {
     pub fn configure(meta: &mut ConstraintSystem<Fp>) -> Self {
-        let username = meta.advice_column();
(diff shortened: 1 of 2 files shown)


sifnoc:
I am sorry to tell you  you that there was a missing part in the benchmark code. It should have been initialized with entry data before performing `full_prover`. I overlooked adding this. Consequently, I believe that the performance results in the verifier are invalid, as the benchmark code lacks a validation check.

You can observe a similar evaluation time between `open_grand_sum` and `open_user_points` with the initialization of entry data, regardless of the unblinded polynomial.

alxkzmn:
Got it. So the difference is only for the empty circuit? Then we can close this PR without merging.

sifnoc:
> Got it. So the difference is only for the empty circuit? Then we can close this PR without merging.

Yes, And I will make PR for fixing bench instead of this.

PR #254, opened 2024-01-11
Implement batched KZG as per FK23

alxkzmn:


kzg_prover/src/utils/batched_kzg.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    // Prepare coefficients vector and zero-pad at the beginning
+    let mut c_fft = vec![Fp::zero(); 2 * d];
+    //Create a reversed copy of the polynomial
+    // let mut f_reversed = f_poly.to_vec();
+    // f_reversed.reverse();
+    c_fft[d..(2 * d)].copy_from_slice(&f_poly);
+
+    println!("c_fft and s_fft assigned");
+    // Assuming omega is a d-th root of unity, and nu is its square (2d-th root of unity)
```
sifnoc, line 46: It seems the "2d-th root of unity" is not omega square.

alxkzmn: At first I didn't believe too, so I wanted to check it. I added
```
println!("nu^2d: {:?}", nu.pow_vartime(&[2 * d as u64]));
```
and the result is 
```
nu^2d: 0x0000000000000000000000000000000000000000000000000000000000000001
```

sifnoc: I understand that $\nu^{2d} = 1$ for a "2d-th root of unity", but I am slightly confused about whether  $\nu^{d}$ should also equal 1. This seems to conflict with the property of $\nu$ being a primitive root, as described in section **1.3 on Discrete Fourier Transform**.

alxkzmn: Got you, it wasn't a primitive root indeed. Fixed it, but the results are still not matching.

kzg_prover/src/circuits/utils.rs
```diff
@@ -198,6 +199,35 @@ pub fn open_user_points(
...
+/// * `Vec<Vec<G1>>` - all KZG opening proofs for the polynomials in the range
+pub fn open_user_points_amortized(
+    advice_polys: &[Polynomial<Fp, Coeff>],
+    params: &ParamsKZG<Bn256>,
+    column_range: Range<usize>,
+    omega: Fp,
+) -> Vec<Vec<G1>> {
+    advice_polys[column_range]
+        // Parallelize the independent amortized openings of the user ID and balance polynomials
+        .par_iter()
```
sifnoc, line 222: I just curious that rayon helps to improve performance in here. because it already `best_fft` well parallelized inside this logic.

kzg_prover/src/utils/amortized_kzg.rs
```diff
@@ -0,0 +1,147 @@
...
+    h.iter_mut().for_each(|h| *h *= n_inv);
+
+    // Truncate to get the first d coefficients
+    h.truncate(d);
+
+    h
+}
+
+/// Compute the naive KZG opening proof
+/// J Thaler, Proofs, Arguments, and Zero-Knowledge, 15.2
```
sifnoc, line 87: I believe you are referring to the book name and chapter. Can you add page numbers to make it look like a book?
Even if it's not just the page number, I think there's a need to tell new comers whether it's a book or not.

kzg_prover/src/utils/amortized_kzg.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    let mut y: Vec<G1> = vec![G1::identity(); 2 * d];
+    y[..d].copy_from_slice(&s_commitments_reversed[..d]);
+
+    // Prepare coefficients vector and zero-pad at the beginning
+    let mut v = vec![Fp::zero(); 2 * d];
+    v[d..(2 * d)].copy_from_slice(f_poly);
+
+    let nu = double_domain.get_omega(); // 2d-th root of unity
+                                        // Perform FFT on s
```
sifnoc, line 62: Is it left over?

kzg_prover/src/utils/amortized_kzg.rs
```diff
@@ -0,0 +1,147 @@
...
+    best_fft(&mut y, nu, (2 * d).trailing_zeros());
+    // Perform FFT on c
```
sifnoc, line 64: Does the "c" refer to "coefficients" of `v`?

sifnoc (changes requested):
I left many quests on comments. Thanks for effort to amortized kzg.
The verifier for amortized KZG is much simple than `VerifierGWC` one.

kzg_prover/src/circuits/tests.rs
```diff
@@ -8,18 +8,124 @@ mod test {
...
+            "Invalid proof verification should fail"
+        );
+
+        // Open the polynomial at the user index (challenge) using the amortized KZG
+
+        // Compute the h vector
+        let mut h = compute_h(&params, f_poly);
+
+        // Demonstrate a single-challenge opening using the calculated h (FK23, eq. 13)
+        let mut challenge_powers = vec![Fp::one(); params.n() as usize];
```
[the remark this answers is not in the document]
alxkzmn: Here we're calculating the powers of challenge $\xi^i$ from 0 to d-1 to calculate $h(X) = ∑h_i\xi^i$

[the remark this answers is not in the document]
alxkzmn: The equation 13 from the paper is referenced in the comment if you want to know more details.

kzg_prover/src/circuits/utils.rs
alxkzmn: Rayon is useful here because we need to calculate multiple openings - one opening for the user ID polynomial and a number of openings for the user balance polynomials, depending on the number of currencies.

kzg_prover/src/utils/amortized_kzg.rs
alxkzmn: Not sure what you're referring to here.

kzg_prover/src/utils/amortized_kzg.rs
alxkzmn: No, it's a naming mismatch between code and paper. The vector before FFT in the paper is named c and the result is named v, but our FFT function modifies the mutable vector in place. I'll unify the naming.

kzg_prover/src/utils/amortized_kzg.rs
sifnoc: I thought the "Perform FFT on s" was left over due to indent.
Now I can see it refers to next line.

kzg_prover/src/utils/amortized_kzg.rs
sifnoc: I see, Thank you.

sifnoc (approved):
We can now accurately locate references (e.g. papers, sections) on code. Thanks.

PR #255, opened 2024-01-21
fix: replace `swap_balances_per_level` with `sum_balances_per_level`

enricobottazzi:
Since we modified the hashing operation inside the Merkle Sum Tree in #188, the `swap_balances_per_level` function has become useless since we no longer have to swap the balances as they are no longer part of the hasher input for the next level. Note that right now we are performing the swap of the balances inside the circuit but the swapped balanced are not consumed anywhere inside the circuit, as you can see by the fact that the swapped balances are not even part of the `return` of `swap_balances_per_level`

https://github.com/summa-dev/summa-solvency/blob/b832e5e2814a7222576a15a8008685983f1beffd/zk_prover/src/chips/merkle_sum_tree.rs#L240-L263

Regarding the performance gains of this modification, the current implementation  `swap_balances_per_level` performs: 

- `bool constraint` and `swap constraint` enabled by `bool_and_swap_selector`
- `sum constraint` enabled by `sum_selector`

The modified implementation of `sum_balances_per_level` only performs `sum constraint`.

Considering a circuit with `n` levels and `m` number of currencies, the modified implementation saves us `2*m*n` constraints. 

Other than that, I updated the verifier contract based on the modified circuit and set `N_BYTES` from 14 to 8 across the repository

Once the PR is reviewed and merged we would need to: 
- update the benchmarks on Gitbook (based on the new circuit and using the updated `N_BYTES`
- update the circuit documentation on Gitbook
- Create a new release tag. 

Also @sifnoc can please check out why the github action is failing?


zk_prover/src/chips/merkle_sum_tree.rs
```diff
@@ -179,86 +179,47 @@ impl<const N_CURRENCIES: usize> MerkleSumTreeChip<N_CURRENCIES> {
...
-
-                // enable the sum_selector at offset 1
-                self.config.sum_selector.enable(&mut region, 1)?;
+                // Extract the values from the cell
+                let current_balance_val = current_balance.value().copied();
+                let element_balance_val = element_balance.value().copied();
 
-                // compute the sum of the two balances and assign it to the column self.config.advice[2] at offset 1
-                let sum = l1_val.zip(r1_val).map(|(a, b)| a + b);
+                // compute the sum of the two balances and assign it to the column self.config.advice[1] at offset 0
```
sifnoc, line 217: typo `config.advice[2]`

sifnoc (changes requested):
Good catch! During my last review, I mistakenly thought that the left and right balance cells were still being used in subsequent rows. Clearly, that's not the case! This modification effectively saves a substantial number of rows, roughly equal to `LEVELS` multiplied by `N_CURRENCIES`!
I left small thing to change, but otherwise, it looks good to me!

sifnoc (approved):

PR #259, opened 2024-01-30
Create solidify verifiers for UnivarateGrandSum and KZG openings 

sifnoc:
Related Issues: 
- [#241](https://github.com/summa-dev/summa-solvency/issues/241)
- [#242](https://github.com/summa-dev/summa-solvency/issues/242)

To enable proof verification on the EVM, the transcript has been switched from `Blake2b` to `Keccak256Transcript`. Additionally, `SingleStrategy` is now applied to the verification of both Snark and KZG proofs, due to the lack of support for `AccumulatorStrategy` in `halo2-solidity-verifier`.

Todos:
- [x] Implement a Solidity verifier for the UnivariateGrandSumCircuit.
- [ ] Implement a Solidity KZG verifier for a proof of grand sum.
- [ ] Implement a Solidity KZG verifier for a proof of user inclusion.

PR #263, opened 2024-02-08
Create solidify verifier for UnivarateGrandSum

sifnoc:
Split from #259:

For testing, the commitment generation, which is `bin/gen_commitment.rs`, can produce a JSON file containing the `snark proof` and `grand sum proof`.

The KZG verifier, which will be able to verify both the grand sum proof and the inclusion proof, will be introduced in an upcoming PR.

- [x] Replace `open_grand_sums` with `create_naive_kzg_proof`.


kzg_prover/benches/kzg.rs
```diff
@@ -8,8 +8,8 @@ use summa_solvency::{
     circuits::{
         univariate_grand_sum::UnivariateGrandSum,
         utils::{
-            full_prover, generate_setup_artifacts, open_grand_sums, open_user_points,
-            open_user_points_amortized, verify_grand_sum_openings, verify_user_inclusion,
```
alxkzmn, line 12: If open_user_points_amortized is deleted, will it break the benchmark?

kzg_prover/bin/gen_verifiers.rs
```diff
@@ -0,0 +1,188 @@
...
+    // Initiate verifier contract
+    let verifier_address = evm.create(verifier_code);
+    let (_, output) = evm.call(verifier_address, proof_calldata);
+
+    // If successfuly verified, the verifier contract will return 1.
+    assert_eq!(output, [vec![0; 31], vec![1]].concat());
+    save_solidity("snark_verifier.sol", &verifier_solidity_fixed);
+    save_solidity("verifying_key.sol", &vk_verifier);
+
+    // 4. Generate Opening Proof for Grand Sums
```
alxkzmn, line 118: I think it's ok to remove the code. related to the grand sum opening until we have an actual verifier for that, otherwise this code acts like a duplicated unit test for the grand sum opening because the result is not used anywhere.

kzg_prover/benches/kzg.rs
sifnoc: Ah, yes. You are right. 
Will fix it.

kzg_prover/bin/gen_verifiers.rs
sifnoc: Okay, I will remove that now.

kzg_prover/src/circuits/utils.rs
```diff
@@ -145,10 +144,33 @@ pub fn open_grand_sums(
...
+        params,
+        &advice_polys[balance_column_range],
+        advice_blinds,
+        challenge,
+        constant_terms,
+    )
+}
+
+// This function for comparison with open_grand_sums, which is used ProverSHPLONK
+pub fn open_grand_sums_gwc(
```
alxkzmn, line 161: Is GWC performing better than SHPLONK?

kzg_prover/src/circuits/utils.rs
```diff
@@ -420,33 +432,31 @@ where
...
+) -> Result<(bool, Vec<Fp>), Box<Error>>
 where
     Scheme::Scalar: WithSmallOrderMulGroup<3>,
 {
-    let mut transcript = T::init(proof);
+    let mut transcript = Keccak256Transcript::new(proof);
 
     // Read the polynomial evaluations from the transcript
-    let evaluations = (0..num_points)
+    let n_points = commitment_points.len();
```
alxkzmn, line 453: Nice catch!

alxkzmn (changes requested):
Thanks for the generators! The only minor changes I would like to request is fixing the benchmark and removing the grand sum-related code from the verifier generator.

kzg_prover/src/circuits/utils.rs
sifnoc: Yes, I updated the result in here: https://hackmd.io/yY8N2nd6RT-giGuGlcrdSA#Benchmark-with-ProverGWC-multi-open-scheme
But we have to benchmark with more than ten currencies. because It has better (or similar performance) in generating proof from UnivarientGrandSum circuit on SHPLONK.

kzg_prover/bin/gen_verifiers.rs
```diff
@@ -0,0 +1,188 @@
...
+    let verifier_solidity_fixed = verifier_solidity
+        .replace("Halo2Verifier", "Verifier")
+        .replace(") public returns (bool)", ") public view returns (bool)");
+
+    let verifier_code = compile_solidity(&verifier_solidity_fixed);
+    let vk_code = compile_solidity(&vk_verifier);
+
+    // 2. Generate Snark Proof for range check
+    //
+    // the instance values has to be at least more than one due to verifier contract that generated from SolidityGenerator.
```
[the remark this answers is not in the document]
alxkzmn: I've created an issue in halo2-solidity-verifier to merge the recent changes https://github.com/summa-dev/halo2-solidity-verifier/issues/1

kzg_prover/bin/gen_verifiers.rs
```diff
@@ -0,0 +1,149 @@
...
+        &[univariate_grand_sum_circuit],
+        &[&[&instances]],
+        &mut OsRng,
+        &mut transcript,
+    );
+    assert!(result.is_ok());
+
+    let result_unwrapped = result.unwrap();
+    result_unwrapped.0.expect("prover should not fail");
+    let _advice_polys = result_unwrapped.1.clone();
```
alxkzmn, line 83: I guess these are the leftovers from the user inclusion verifier, could you please delete them?

sifnoc: It removed!

Issue #245, opened 2023-12-28
V2 Docs


enricobottazzi:
Add reference to: 

#237 
#236 
#235 

Structure of the docs 
1.1 Summa
1.2 Protocol Flow

2.1 Commitments
2.2 zkSNARKs
2.3 Kzg Commitments 
2.4 Univariate Polynomial GrandSum

3.1 Summa Solvency

4.1 UnivariateGrandSum Circuit
4.2 Halo2 Utils 

5.1 Summa.sol Smart Contract

- [x] Add reference to correct release to V1

alxkzmn:
Additional things to mention in the docs:

- [ ] Summa contract deployer should modify the expected zk-proof length in the assertion for the specific circuit configuration (see discussion here: https://github.com/summa-dev/summa-solvency/pull/264/files#r1489028964)

sifnoc:
This can be closed due to release v2 doc on gitbook.

PR #264, opened 2024-02-14
Refactor contract to use the univariate grand sum zkSNARK verifier

alxkzmn:


sifnoc (changes requested):

contracts/src/Summa.sol
```diff
@@ -11,12 +11,10 @@ import "./interfaces/IVerifier.sol";
 contract Summa is Ownable {
     /**
      * @dev Struct representing the configuration of the Summa instance
-     * @param mstLevels The number of levels of the Merkle sum tree
-     * @param currenciesCount The number of cryptocurrencies supported by the Merkle sum tree
-     * @param balanceByteRange The number of bytes used to represent the balance of a cryptocurrency in the Merkle sum tree
+     * @param currenciesCount The number of cryptocurrency balances encoded in the polynomials
+     * @param balanceByteRange The number of bytes used to represent the balance of a cryptocurrency in the polynomials
```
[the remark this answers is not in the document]
alxkzmn: Actually it wasn't used in the previous version of the contract. The idea was to force the Custodian to publicly commit to a value just for the record.

contracts/src/Summa.sol
```diff
@@ -46,63 +44,72 @@ contract Summa is Ownable {
...
+            _verificationKey != address(0),
+            "Invalid verifying key address"
+        );
+        verificationKey = _verificationKey;
+        require(
+            address(_polynomialEncodingVerifier) != address(0),
+            "Invalid polynomial encoding verifier address"
+        );
+        polynomialEncodingVerifier = _polynomialEncodingVerifier;
+        config = SummaConfig(currenciesCount, balanceByteRange);
```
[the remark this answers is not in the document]
alxkzmn: I will try to take advantage of that, thank you! If there is a straightforward way of calculating it, we can always calculate this value from the VK and not use any variables.

contracts/src/Summa.sol
```diff
@@ -136,22 +143,36 @@ contract Summa is Ownable {
...
      */
     function submitCommitment(
-        uint256 mstRoot,
-        uint256[] memory rootBalances,
+        bytes memory proof,
         Cryptocurrency[] memory cryptocurrencies,
         uint256 timestamp
     ) public onlyOwner {
-        require(mstRoot != 0, "Invalid MST root");
+        require(proof.length == 5376, "Invalid proof length");
```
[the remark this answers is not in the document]
alxkzmn: I think we can still have a fixed value in the assertion and we can mention it in the docs. Eventually, the contract will be deployed for the specific circuit configuration whose proof length will stay the same. I will reference this in the issue for V2 docs.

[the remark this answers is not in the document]
alxkzmn: https://github.com/summa-dev/summa-solvency/issues/245#issuecomment-1943390536

contracts/src/interfaces/IVerifier.sol
```diff
@@ -7,12 +7,14 @@ pragma solidity ^0.8.18;
...
-     * @param pubInputs The public inputs to the proof
+     * @param vk The verification key
      * @param proof The proof
+     * @param instances The public inputs to the proof
      * @return true if the proof is valid, false otherwise
      */
-    function verify(
-        uint256[] calldata pubInputs,
-        bytes calldata proof
+    function verifyProof(
```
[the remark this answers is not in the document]
alxkzmn: A non-view function can call a view function without any issues, so there's no need to modify the inner view function.

alxkzmn:
I've implemented the check that the number of cryptocurrencies corresponds to the verification key. The method is explained in the comments, but I need to add that I have tried to read the full value of the last permutation and the empty slot after it which lead the transaction to run out of gas. This is the reason why I'm only reading the 32 bytes that overlap the last value and the empty slot behind it by using a single `extcodecopy`.

contracts/test/Summa.ts
```diff
@@ -61,6 +48,129 @@ describe("Summa Contract", () => {
...
+        "src/VerifyingKey.sol:Halo2VerifyingKey"
+      );
+      await verifyingKey.deployed();
+      await expect(
+        ethers.deployContract("Summa", [
+          verifyingKey.address,
+          ethers.constants.AddressZero,
+          ["ETH", "BTC"],
+          ["ETH", "BTC"],
+          0, // Invalid byte range
```
sifnoc, line 127: Regardless the revert message, should the 'balance byte range' parameter be explicitly set to 0 in other test cases?

contracts/test/Summa.ts
```diff
@@ -47,8 +33,9 @@ describe("Summa Contract", () => {
     const summa = await ethers.deployContract("Summa", [
       verifyingKey.address,
       snarkVerifier.address,
-      2, // The number of cryptocurrencies in the balance polynomials
-      14, // The number of bytes used to represent the balance of a cryptocurrency in the polynomials
+      ["ETH", "BTC"],
+      ["ETH", "BTC"],
+      8, // The number of bytes used to represent the balance of a cryptocurrency in the polynomials
```
sifnoc, line 38: I believe this number refers to the number of advice columns in the range check chip in V1. Please correct me if it is wrong.
However, its relevance appears vague to me in V2.
Could you provide a bit more explanation about this?

contracts/src/Summa.sol
```diff
@@ -77,27 +57,89 @@ contract Summa is Ownable {
...
+     */
+    function validateVKPermutationsLength(
+        address vkContract,
+        uint256 numberOfCurrencies
+    ) internal view returns (bool isValid) {
+        // The number of permutations is 2 + 4 * numberOfCurrencies because of the circuit structure:
+        // 1 per instance column, 1 per constant column (range check) and 4 per range check columns times the number of currencies
+        uint256 numPermutations = 2 + 4 * numberOfCurrencies;
+
+        uint256 startOffsetForPermutations = 0x2e0; // The value can be observed in the VerificationKey contract, the offset is pointing after all the parameters and the fixed column commitment
```
sifnoc, line 120: For record, this position may changed at recent `halo2-solidity-verifier`
related issue: https://github.com/summa-dev/halo2-solidity-verifier/issues/1

contracts/src/Summa.sol
```diff
@@ -77,27 +57,89 @@ contract Summa is Ownable {
...
+            numPermutations *
+            0x40;
+
+        // extcodecopy is a gas-expensive operation per byte, so we want to minimize the number of bytes we read.
+        // This hack is to read the 32 (0x20) bytes that overlap the last permutation and the empty memory location behind it.
+        // For example, a circuit with 2 currencies will have 10 permutations, so the location behind the last permutation will be at 0x2e0 + 10 * 0x40 = 0x0560
+        // We read 0x20 bytes starting from 0x0550, which will include a piece of the last permutation and the empty memory location behind it.
+        uint256 readOffset = offsetAfterLastPermutation - 0x10;
+        bool valid;
+        // Extract the last permutation
```
sifnoc, line 133: Nice!

sifnoc (changes requested):
Overall, It looks good to me!

contracts/test/Summa.ts
alxkzmn: Do you mean that we're not using this value for any checks inside the contract?

contracts/test/Summa.ts
alxkzmn: It shouldn't, thanks for pointing out!

contracts/test/Summa.ts
sifnoc: > Do you mean that we're not using this value for any checks inside the contract?

No, I understood this since your previous comment https://github.com/summa-dev/summa-solvency/pull/264#discussion_r1489167346

I had understood this number as representing the number of cells used in the range check circuit. which is now 4 in V2.
 However, I was a bit confused, and you're correct. It is indeed 64 bits (8 bytes). I made a mistake by calculating it in my mind. Sorry for bringing this up.

I would just like to mention here, in the comment, that the term 'balanceByteRange' signifies the maximum balance that is safe from overflow.

sifnoc (approved):

contracts/test/Summa.ts
alxkzmn: Added a constructor NatSpec, also added the check of the permutations number according to balance byte range.

sifnoc: Thanks for updating!

PR #265, opened 2024-02-15
fix: backend and test on backend

sifnoc:
- I've temporarily commented out lines that are not fully implemented yet.
- Additionally, I've marked sections that require refactoring with TODO comments.
- This adjustment allows the backend to test the generation of inclusion proofs using the create_naive_kzg_proof method.

### DIFF
--- a/backend/examples/summa_solvency_flow.rs
+++ b/backend/examples/summa_solvency_flow.rs
@@ -74,7 +74,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let univariate_grand_sum_circuit =
         UnivariateGrandSum::<N_USERS, N_CURRENCIES>::init(entries.to_vec());
 
-    let (params, pk, _) = generate_setup_artifacts(K, None, &univariate_grand_sum_circuit).unwrap();
+    let (params, pk, vk) =
+        generate_setup_artifacts(K, None, &univariate_grand_sum_circuit).unwrap();
 
     // Create a proof
     let instances = vec![Fp::one(); 1]; // This instance is necessary to verify proof on solidity verifier.
@@ -88,7 +89,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Using the `round` instance, the commitment is dispatched to the Summa contract with the `dispatch_commitment` method.
     let timestamp = 1u64;
     let mut round =
-        Round::<N_CURRENCIES, N_POINTS, N_USERS>::new(&signer, advice_polys, params_path, 1)
+        Round::<N_CURRENCIES, N_POINTS, N_USERS>::new(&signer, advice_polys, params, vk, 1)
             .unwrap();
 
     // // Sends the commitment, which should ideally complete without errors.
--- a/backend/src/apis/mod.rs
+++ b/backend/src/apis/mod.rs
@@ -1,8 +1,3 @@
 pub mod address_ownership;
 pub mod csv_parser;
 pub mod round;
-
-use ethers::types::U256;
-use num_bigint::BigUint;
-use num_traits::Num;
-use summa_solvency::entry::Entry;
--- a/backend/src/apis/round.rs
+++ b/backend/src/apis/round.rs
@@ -2,10 +2,10 @@ use ethers::types::{Bytes, U256};
 use halo2_proofs::{
     arithmetic::{best_fft, Field},
     halo2curves::{
-        bn256::{Bn256, Fr as Fp, G1Affine},
+        bn256::{Bn256, G1Affine},
         group::Curve,
     },
-    plonk::{AdviceSingle, ProvingKey, VerifyingKey},
+    plonk::{AdviceSingle, VerifyingKey},
     poly::{
         kzg::commitment::{KZGCommitmentScheme, ParamsKZG},
         Coeff,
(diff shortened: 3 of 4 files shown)


backend/examples/summa_solvency_flow.rs
```diff
@@ -57,23 +65,56 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    let (zk_snark_proof, advice_polys, _omega) = full_prover(
+        &params,
+        &pk,
+        univariate_grand_sum_circuit.clone(),
+        &[instances.clone()],
+    );
+
+    // Iterate entries and collect values to `[Vec[Fp]; N_POINTS]`
+    const FP_ARRAY: Vec<Fp> = Vec::new();
+    let mut user_data = [FP_ARRAY; N_POINTS];
```
alxkzmn, line 90: I think we can use the entries directly without user_data because user_data array has the same format.

backend/src/apis/mod.rs
```diff
@@ -23,6 +23,7 @@ where
     let entry: Entry<N_CURRENCIES> = Entry::new(username, balances.try_into().unwrap()).unwrap();
 
     // Convert Fp to U256
-    let hash_str = format!("{:?}", entry.compute_leaf().hash);
-    U256::from_str_radix(&hash_str, 16).unwrap()
+    // TODO: Instead of hash of userdata, What should we out to this function?
```
alxkzmn, line 26: Could you please clarify this TODO? I think we can just remove the functionality related to the MST.

backend/src/apis/round.rs
```diff
@@ -38,35 +34,48 @@ impl MstInclusionProof {
...
-pub struct Round<'a, const LEVELS: usize, const N_CURRENCIES: usize, const N_BYTES: usize> {
+/// The `Round` struct represents a single operational cycle within the Summa Solvency protocol.
+///
+/// # Type Parameters
+///
+/// * `N_CURRENCIES`: The number of currencies for which solvency is verified in this round.
+/// * `N_POINTS`: The number of points in the `UnivariateGrandSum` circuit, which is `N_CURRENCIES + 1`.
+/// * `N_USERS`: The number of users involved in this round of the protocol.
+///
+/// /// These parameters are used for initializing the `UniVariantGrandSum` circuit within the `Snapshot` struct.
```
alxkzmn, line 45: UnivariateGrandSum

backend/src/apis/round.rs
```diff
@@ -75,100 +84,116 @@ where
...
 {
     pub fn new(
-        mst: Box<dyn Tree<N_CURRENCIES, N_BYTES>>,
+        advice_polys: AdviceSingle<G1Affine, Coeff>,
+        user_balances: [Vec<Fp>; N_POINTS],
         params_path: &str,
-    ) -> Result<Snapshot<LEVELS, N_CURRENCIES, N_BYTES>, Box<dyn std::error::Error>> {
-        let mst_inclusion_circuit = MstInclusionCircuit::<LEVELS, N_CURRENCIES, N_BYTES>::init_empty();
+    ) -> Result<Snapshot<N_CURRENCIES, N_POINTS, N_USERS>, Box<dyn Error>> {
+        let univariant_grand_sum_circuit: UnivariateGrandSum<N_USERS, N_CURRENCIES> =
```
alxkzmn, line 133: univariate_grand_sum_circuit

backend/src/apis/round.rs
```diff
@@ -75,100 +84,116 @@ where
...
 
         // get k from ptau file name
-        let parts: Vec<&str> = params_path.split("-").collect();
+        let parts: Vec<&str> = params_path.split('-').collect();
         let last_part = parts.last().unwrap();
         let k = last_part.parse::<u32>().unwrap();
 
-        let mst_inclusion_setup_artifacts: SetupArtifacts =
-            generate_setup_artifacts(k, Some(params_path), mst_inclusion_circuit).unwrap();
+        let univariant_grand_sum_setup_artifcats =
```
alxkzmn, line 141: univariate_grand_sum_setup_artifacts

backend/src/apis/mod.rs
sifnoc: I think it's left over in previous commit.

backend/src/tests.rs
```diff
@@ -217,81 +256,120 @@ mod test {
...
+        }
 
-        // After sending transaction of liability commitment, logs should be updated
-        liability_commitment_logs = summa_contract
-            .liabilities_commitment_submitted_filter()
-            .query()
-            .await?;
+        let mut round = Round::<N_CURRENCIES, N_POINTS, N_USERS>::new(
+            &signer,
+            advice_polys,
```
alxkzmn, line 310: advice_polys and user_data are representing exactly the same data in a slightly different way, so we're effectively duplicating the data. for a large user base this would cost a lot of memory. I think we need to have only one of the representations stored inside the round.

alxkzmn (changes requested):
I think it's necessary to decide whether we're storing the advice_polys or user_data in the round.

sifnoc:
> I think it's necessary to decide whether we're storing the advice_polys or user_data in the round.

For generating inclusion proof, we can evaluate `z` with `advice_poly` instead of using `entries`. Thanks for point out!

backend/src/apis/round.rs
```diff
@@ -75,100 +85,119 @@ where
...
+            generate_setup_artifacts(k, Some(params_path), &univariate_grand_sum_circuit).unwrap();
```
alxkzmn, line 143: If we already have the polynomials at this step, it means that we have obtained the range check zkSNARK proof that used some setup, I think we can reuse the same setup here by also passing it as an argument.

sifnoc: Oh, yes! you are right. I will fix soon.

alxkzmn (approved):
LGTM

PR #269, opened 2024-02-26
refactor: simple code cleanups in mst implementation

teddav:
Just cleaned a few things while reviewing the MST. All tests are passing, but some changes (ex: removing some condition) might require to add more tests. Let me know if you think that's the case


alxkzmn:
@ntampakas is it possible to let the workflows from forks to run and use the AWS credentials? Or is there a better way to deal with external contributions?

sifnoc (approved):

ntampakas:
> @ntampakas is it possible to let the workflows from forks to run and use the AWS credentials? Or is there a better way to deal with external contributions?

Hey Alex!
Probably you are referring to benchmark workflow that uses a self-hosted runner. Is this correct?
AFAIN, a forked repo, cannot execute workflows on the upstream side. What is the exact use case?

alxkzmn:
@ntampakas I'm referring to the Rust workflow for tests that runs on push, it is also using the self-hosted runner because of the past performance issues with the github runner.

teddav:
I just talked with @sifnoc who explained to me why you moved to a self-hosted runner. Seems the change was made in PR #199 and time went from 15' to about 4'. I checked [the runs before the PR](https://github.com/summa-dev/summa-solvency/actions/workflows/rust.yml?page=18) and they were indeed around 15min but unfortunately I can't see the details because Github doesn't keep that information for that long, so I don't know what part was slow. But looking at more recent ones, it seems like running is actual test is the slow part, while installing the different components is fast. So my suggestion: why not just split the tests in different jobs (which will be run in parallel). Currently they are all run into 1 big job called "build". And I see 4 tasks being run:
- Test Zk Prover
- Test Zk Prover examples
- Test backend
- Test backend example
Let's split them and see. If it's unequal we can figure out how to split them differently.

teddav:
i just pushed the change. i created a new workflow. I guess we can keep both running in parallel for now just to check the difference, and then decide which one to keep. We could also cache dependencies to make things even faster. I dont know if caching is possible for rust build/rust test though

alxkzmn:
How about switching between self-hosted runner and GitHub runner based on where the changes are coming from? E.g., if it's a fork, run in GitHub, but if it's in Summa org, use the current setup? Here's the way to detect the repo name: https://github.com/orgs/community/discussions/25217#discussioncomment-3246901

.github/workflows/tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Install solc
+        run: (hash svm 2>/dev/null || cargo install --version 0.2.23 svm-rs) && svm install 0.8.20 && solc --version
+
+      - name: Test Zk Prover examples
+        run: |
+          cd zk_prover
+          cargo run --release --example gen_inclusion_verifier
+          cargo run --release --example gen_commitment
+          cargo run --release --example gen_inclusion_proof
+          cargo run --release --example nova_incremental_verifier
```
sifnoc, line 42: We can run the example `nova_incremental_verifier` in a new job without `installing solc` step.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,70 @@
...
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  test-zk-prover:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install solc
```
sifnoc, line 21: In this job, we can remove step `install solc` because the test is not require`solc`.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,70 @@
...
+          cargo run --release --example gen_inclusion_verifier
+          cargo run --release --example gen_commitment
+          cargo run --release --example gen_inclusion_proof
+          cargo run --release --example nova_incremental_verifier
+
+  test-backend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install solc
```
sifnoc, line 48: In here also, not require `install solc` step.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Test backend
+        run: |
+          cd backend
+          cargo test --release -- --nocapture
+
+  test-backend-examples:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install solc
```
sifnoc, line 62: In here also, not require `install solc` step.

sifnoc (changes requested):
Thank you for the initiative to optimize the github action flow.

We had a specific reason for running tests sequentially. It's because the interface file for the verifier contract in the Backend, specifically `backend/src/contracts/generated/inclusion_verifier.rs`, is generated from the ABI file, which in turn is generated from the contract. This process is similar to what is done in the script `update_verifier_contract.sh` in the Backend.

However, now that the circuits and contract have matured for V1, your suggestion to test in a parallelized manner seems very appropriate.

Also, following Alexs` found, we can add condition to the workflow file that run test on self-host runner.

sifnoc (approved):
Thank you for fixing this. Now, the test finishes in about 6 minutes without using the est instance.

.github/workflows/tests.yml
alxkzmn, line 1: Should this new workflow be merged with the old one?

alxkzmn (changes requested):
Let's figure out which testing workflow we're using before we merge this PR. If the parallel tests work well, should they be merged into the old workflow?

.github/workflows/tests.yml
teddav: I think we can. It seems like it's working as expected
cc @sifnoc

teddav: we can also adapt it for v2. When are you planning on merging v2 to main btw?

sifnoc: > I think we can. It seems like it's working as expected cc @sifnoc

It would be great to combine the two files into one file (`rust.yml`). Actually, I couldn't thought about that is possible.

sifnoc: > we can also adapt it for v2. When are you planning on merging v2 to main btw?

We had decided to leave it v2 branch as it is instead of merge it to main for now.
So, there is no plan to merge it to main right now.

alxkzmn:
@teddav Do I get it right that now the tests would either run on both AWS and GitHub when pushed from Summa org or only on GitHub when pushed from a fork? In case of Summa org I was expecting the tests to only run in AWS, not on GitHub.

teddav:
@alxkzmn yes you're right. It's useless to run on both. Even though now the AWS worker seems a bit overkill even for Summa's team. It seems like everything could run fine just on Github

sifnoc:
This PR has been thoroughly discussed and is ready for merging. I appreciate the effort put into this PR, @teddav . 

Based on the discussions, it seems that the next step for this PR would be to remove the requirement of an AWS instance in the workflow.

PR #273, opened 2024-03-07
Created dummy entry generator for benchmark

sifnoc:
- The current benchmark is limited by its capacity to handle polynomials for a relatively small number of users, specifically generating proofs for only 16 or 64 users in its benchmarking code. To address this limitation, it is necessary to dynamically initialize entries based on the specified number of users, `N_USERS`, instead of relying on entry data from a CSV file, which constrains us to a fixed user count.
- Fixed the number of user, `N_USERS` in the benchmark code, `kzg.rs` with maximum number of user within given `K`   
- Also, Replaced `N_ASSETS` to `N_CURRENCIES`

### DIFF
--- a/kzg_prover/benches/kzg.rs
+++ b/kzg_prover/benches/kzg.rs
@@ -20,7 +20,7 @@ use summa_solvency::{
     },
     cryptocurrency::Cryptocurrency,
     entry::Entry,
-    utils::{big_uint_to_fp, parse_csv_to_entries},
+    utils::{big_uint_to_fp, generate_dummy_entries},
 };
 
 fn bench_kzg<
@@ -31,7 +31,6 @@ fn bench_kzg<
     CONFIG: CircuitConfig<N_CURRENCIES, N_USERS>,
 >(
     name: &str,
-    csv_path: &str,
 ) where
     [(); N_CURRENCIES + 1]:,
 {
@@ -59,7 +58,7 @@ fn bench_kzg<
 
     let mut entries: Vec<Entry<N_CURRENCIES>> = vec![Entry::init_empty(); N_USERS];
     let mut cryptos = vec![Cryptocurrency::init_empty(); N_CURRENCIES];
-    parse_csv_to_entries::<&str, N_CURRENCIES>(csv_path, &mut entries, &mut cryptos).unwrap();
+    let _ = generate_dummy_entries(&mut entries, &mut cryptos);
 
     // Calculate total for all entry columns
     let mut csv_total: Vec<BigUint> = vec![BigUint::from(0u32); N_CURRENCIES];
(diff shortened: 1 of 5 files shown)


kzg_prover/benches/kzg.rs
```diff
@@ -261,75 +260,65 @@ fn bench_kzg<
...
-    const N_POINTS: usize = 3;
+    const N_CURRENCIES: usize = 1;
+    const N_POINTS: usize = 2;
 
     // Demonstrating that a higher value of K has a more significant impact on benchmark performance than the number of users
     #[cfg(not(feature = "no_range_check"))]
     {
         const K: u32 = 18;
-        const N_USERS: usize = 16;
+        const N_USERS: usize = (1 << 17) + (1 << 16) - 6;
```
alxkzmn, line 270: Can you try to replace it with 2^K-6?

kzg_prover/src/utils/dummy_entries.rs
```diff
@@ -0,0 +1,102 @@
...
+use rand::{distributions::Alphanumeric, Rng};
+use rayon::prelude::*;
+use std::error::Error;
+
+use crate::cryptocurrency::Cryptocurrency;
+use crate::entry::Entry;
+
+// This is for testing purposes with a large dataset instead of using a CSV file
+pub fn generate_dummy_entries<const N_CURRENCIES: usize>(
+    entries: &mut [Entry<N_CURRENCIES>],
```
alxkzmn, line 11: In this case it should be ok to return the (entries, currencies) from the function instead of passing in and filling the mutable arrays (which is used in parser to save the memory).

alxkzmn (changes requested):

kzg_prover/benches/kzg.rs
sifnoc: Indeed, that is worked. Similarly, we can apply this to `N_POINTS` as follows: 
```Rust
const N_POINTS = N_CURRENCIES + 1;
```

kzg_prover/benches/kzg.rs
```diff
@@ -59,7 +58,7 @@ fn bench_kzg<
 
     let mut entries: Vec<Entry<N_CURRENCIES>> = vec![Entry::init_empty(); N_USERS];
     let mut cryptos = vec![Cryptocurrency::init_empty(); N_CURRENCIES];
-    parse_csv_to_entries::<&str, N_CURRENCIES>(csv_path, &mut entries, &mut cryptos).unwrap();
+    let _ = generate_dummy_entries(&mut entries, &mut cryptos);
```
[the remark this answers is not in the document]
sifnoc: It is for warning, but this one would be fix with for https://github.com/summa-dev/summa-solvency/pull/273#discussion_r1517360553

kzg_prover/benches/kzg.rs
```diff
@@ -261,75 +258,65 @@ fn bench_kzg<
...
-    const N_POINTS: usize = 3;
+    const N_CURRENCIES: usize = 1;
+    const N_POINTS: usize = N_CURRENCIES + 1;
 
     // Demonstrating that a higher value of K has a more significant impact on benchmark performance than the number of users
     #[cfg(not(feature = "no_range_check"))]
     {
         const K: u32 = 18;
-        const N_USERS: usize = 16;
+        const N_USERS: usize = 2usize.pow(K - 1) + 2usize.pow(K - 2) - 6; // Which is equal to (1 << 17) + (1 << 16) - 6
```
alxkzmn, line 268: Can you explain the value 2^(K-1)-2^(K-2)-6? Why in the case of K=17 you're using 2^(K-1)-6?

alxkzmn: I think the correct max value should be 2^K - 6.

alxkzmn (changes requested):
I think the correct max value for the number of users in the circuit should be 2^K-6.

kzg_prover/benches/kzg.rs
sifnoc: I'm unable to provide a clear explanation for the failure of the $2^K - 6$ benchmark. Attempting to run it resulted in an error: 'vk generation should not fail: NotEnoughRowsAvailable { current_k: 17 }'.

sifnoc: This has led me to believe that the maximum value for `N_USER` might be within the range between $2^{(K - 1)}$ and $2^K - 6`.

PR #275, opened 2024-03-14
Remove the unused N_POINTS generic constant

alxkzmn:
The N_POINTS was originally used in `utils.rs` but got replaced with a range, so it hasn't been used for awhile.


sifnoc (approved):
Thanks! It much clear now!

Issue #274, opened 2024-03-12
Find out the max number of entries that is possible to put into the advidce columns given a certain K


alxkzmn:
The expected maximum number of usable columns for the univariate grand sum circuit is 2^K-6.
https://github.com/summa-dev/halo2/blob/main/halo2_proofs/src/plonk/circuit.rs#L2349
https://github.com/summa-dev/halo2/blob/main/halo2_proofs/src/plonk/prover.rs#L310
However, the observed number is much lower.

sifnoc:
The current maximum number of users that can be accommodated is $2^K - 2^{16} - 6$ in the advice column, given K. The deduction of 6 rows for blind factors, and $2^{16}$ rows represents the number of values occupied in the fixed column. These are constant values ranging from 0 to $2^{16}$, utilized for the range check chip. This configuration leads to an issue where the UnivariateGrandSum circuit cannot be fully populated with $2^K - 6$ entries in the advice columns, despite there being $2^{16}$ empty rows in these columns.

The table structure is illustrated as follows:

| Number of Row  | Username | Currency_1 | Currency_2 | ... | range_u16 |
|----------------|----------|------------|------------|-----|-----------|
| 1              | dxGaEAii | 11888      | 41163      |     | 0         |
| ...            |          |            |            |     |           |
| 65536          | MBlfbBGI | 67823      | 18651      |     | 65535     |
| 65537          | lAhWlEWZ | 18651      | 2087       |     | 0         |
| 65538          | nuZweYtO | 22073      | 55683      |     | 0         |
| ...            |          |            |            |     | ...       |
| $2^K - 2^{16} - 6$ | RZNneNuP | 31699      | 35479      |     | 0         |
| $2^K - 2^{16} - 5$ |          |            |            |     | 0         |
| ...            |          |            |            |     | ...       |
| 2^K            |          |            |            | ... | 0         |

The root cause of this issue stems from the `constraint_constant` method, which is used to verify the zero value in the range check chip. This method performs two primary operations: it assigns a constant value to the fixed column (denoted as `range_u16` in the table) and then mapping this value(internally called, `permutation.copy`) to the advice column. Essentially, the function assigns values to the next available row after the last used row each time it is called, meaning it populates the next empty row in the fixed column.

At row $2^K - 2^{16} - 5$, the range check chip cannot assign 0 values to `range_u16` anymore, as the column is already filled with other values.

Below is the actual trace log for reference:
```
[...]

PR #276, opened 2024-03-14
Fix maximum entries in polynomial interpolation approach

sifnoc:
To address the issue #274 

- Added advice column in the range check circuit
- Assign zero value from instance value
- Replace terms `polynomial encoding` with `polynomial interpolation`


kzg_prover/src/chips/range/tests.rs
```diff
@@ -130,6 +131,12 @@ impl Circuit<Fp> for TestCircuit {
 
         let add_selector = meta.selector();
 
+        let instance = meta.instance_column();
+        meta.enable_equality(instance);
+
+        let z0 = meta.advice_column();
```
alxkzmn, line 137: Seems like z0 is not used anymore

sifnoc: Thanks, I quickly fixed! at [c831f5d](https://github.com/summa-dev/summa-solvency/pull/276/commits/c831f5d2f8e28adb0eed0de593eb3c93ddc0ecd7)

alxkzmn (changes requested):
Looks good, you can merge after removing the unused z0 advice column. Thanks for cleaning up the naming!

alxkzmn (approved):
LGTM

PR #278, opened 2024-03-19
Add range check sufficiency explanation (resolves #237)

### LINKED ISSUE #237: Risk of overflow in the grandsum
All the coefficients of the polynomial from which we eventually extract the grandsum are defined modulo the prime field of the circuit. The constant term itself, which is later multiplied by the degree of the polynomial to obtain the grandsum is a value modulo the prime field of the circuit. Because of that, it is possible that the grandsum overflows the prime field of the circuit wraps back to 0. 

This undesired behavior is simulated in this [test](https://github.com/summa-dev/summa-solvency/blob/v2-grandsum-overflow-test/kzg_prover/src/circuits/tests.rs#L142-L241). In particular, we use [input balances](https://github.com/summa-dev/summa-solvency/blob/v2-grandsum-overflow-test/csv/entry_16_prime_overflow.csv) which, when summed, will overflow the prime field of the circuit. Furthermore, in order to simulate this behaviour we remove the range check from the `MstInclusionCircuit`.

As expected, the opening grandsum extracted from the polynomial commitment doesn't match the actual sum from the csv file because of the overflow.
https://github.com/summa-dev/summa-solvency/blob/d3adb15ecb2b82ee6146afc88bc15543af8e6c22/kzg_prover/src/circuits/tests.rs#L210-L212

So, the real question is whether there's a risk of overflow in the grandsum in production when a 64 bits range check is performed on each entry balance. Remember that the prime field of the circuit is 254 bits. Let's consider the case at the limit in which we have 2^28 users and all the balances are the maximum possible (namely 2^64 - 1).

```
>>> 2**28 * (2**64-1)
4951760157141521099328061440
>>> n = 4951760157141521099328061440
>>> num_bits = n.bit_length()
>>> print(num_bits)
92
```

The resulting grand sum is only 92 bits. Therefore, we can conclude that using the grandsum together with a range check of 64 bits on the input balances, safely removes the risk of overflow in the grandsum calculation.

alxkzmn:
Resolves #237.

### DIFF
--- a/kzg_prover/src/chips/range/range_check.rs
+++ b/kzg_prover/src/chips/range/range_check.rs
@@ -9,6 +9,21 @@ use std::fmt::Debug;
 /// Configuration for the Range Check u64 Chip
 /// Used to verify that an element lies in the u64 range.
 ///
+/// To prove that the 64-bit balance values would not cause the overflow of the grand sum,
+/// let's consider the case at the limit in which we have 2^28 users and all their
+/// balances are the maximum possible (namely, 2^64-1):
+///
+/// >>> 2**28 * (2**64-1)
+/// 4951760157141521099328061440
+/// >>> n = 4951760157141521099328061440
+/// >>> num_bits = n.bit_length()
+/// >>> print(num_bits)
+/// 92
+///
+/// The resulting grand sum is only 92 bits. Therefore, we can conclude that the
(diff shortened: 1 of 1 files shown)


sifnoc (approved):
Looks good to me! Let's merge it.

PR #281, opened 2024-03-26
add isNotOnlyWhitespace check

Einstellung:
I think just check `signature` and `message` length is insufficient. If user input data with just blank space then the check will  fail. Therefore I added an `isNotOnlyWhitespace` function to enhance the checking capability

### DIFF
--- a/contracts/src/Summa.sol
+++ b/contracts/src/Summa.sol
@@ -209,8 +209,8 @@ contract Summa is Ownable {
             require(
                 bytes(_addressOwnershipProofs[i].cexAddress).length != 0 &&
                     bytes(_addressOwnershipProofs[i].chain).length != 0 &&
-                    _addressOwnershipProofs[i].signature.length != 0 &&
-                    _addressOwnershipProofs[i].message.length != 0,
+                    isNotOnlyWhitespace(_addressOwnershipProofs[i].signature) &&
+                    isNotOnlyWhitespace(_addressOwnershipProofs[i].message),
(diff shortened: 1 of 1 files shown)


sifnoc:
Thank you for your contribution and the effort you put into enhancing the `submitProofOfAddressOwnership` function. We appreciate your attention to detail and your suggestion to prevent potential issues with whitespace in signatures and messages.

We have intentionally chosen to keep the assertions minimal for the ownership proof function. Our decision is based on the assumption that Custodians are trusted entities and would not attempt to include malicious signatures or messages. If a Custodian were to submit an odd ownership proof, it would be easily noticeable on the contract and could potentially damage their reputation.

The additional `isNotOnlyWhitespace` check introduces an increased gas cost of approximately 2000 gas for the submitProofOfAddressOwnership function in testing. The gas consumption would be even higher if the message is longer. For efficient Summa operation, we strive to minimize operating costs, and this extra gas consumption would have a direct impact on Summa's operational cost.

Considering our current priorities and the trust we place in Custodians, we believe that the added `isNotOnlyWhitespace` feature may not be the best fit for the Summa contract at this time.
However, if you have any ideas on how to validate the message and signature more efficiently than the current implementation, we would be more than happy to review and consider your suggestions.
We appreciate your understanding and look forward to any future contributions you may have.

Einstellung:
You are right, it seems `isNotOnlyWhitespace` is not solve enough problems but cost extra gas. Now I will close the comment, when I have some other new idea then I will pull request again.

PR #282, opened 2024-03-26
Remove N_BYTES generic parameter form Tree trait

### LINKED ISSUE #256: Remove `N_BYTES` as generic parameter from the `Tree` trait
From what I can tell, it is useless. As I remember correctly, it was previously used to warn the tree builder for a risk of a balance overflow during tree generation, but we removed that check.

alxkzmn:
Resolves #256 .


sifnoc:
Looks good.
I've identified some places where the N_BYTES value needs updating. I will change it from 14 to 8 once this is merged.

sifnoc (approved):

PR #284, opened 2024-03-28
Optimize verifiers

sifnoc:
There are unused functions in assembly block in verifier contract.

Removed them for saving deploy gas cost.


alxkzmn (approved):
Great!

PR #292, opened 2024-05-13
Apply sumcheck to nonzero constraints directly

alxkzmn:
Please see https://github.com/summa-dev/plonkish/pull/2 to understand how the ["nonzero" gate](https://github.com/summa-dev/summa-solvency/blob/2460efe324bbbf000e97e2c254990bb6321b84cd/prover/src/circuits/summa_circuit.rs#L92-L99) is handled in the modified plonkish backend. TLDR: 
- we put a negative grand total below the balances in the same column;
- we create a "nonzero" gate that's simply querying the current cell;
- the modified plonkish backend sums them all up when including into sumcheck, so if we have a circuit that looks like this

|Col|Balance|
|-|-|
|Row0|3|
|Row1|5|
|Row2|-8|

a simple gate that is querying the balance like `meta.query_advice(balance, Rotation::cur())` will add all the cells to sumcheck without the need for rotations, and the final sumcheck (which is actually a zero check in hyperplonk) will pass (`3 + 5 -8 =0`).


sifnoc:
Looks good! Let's merge this.
After then I will rename this branch name from "v3" to "v3b".

PR #293, opened 2024-06-03
Added concatenated balance for nonzero constraint process

sifnoc:
I added an additional advice column for the concatenated balance, which combines more than two balances into one with shifted positions. For example, consider the following total balances:
- (A) 0x0000000000000000000000000000000000000000000000000000000187094dba
- (B) 0x00000000000000000000000000000000000000000000000000000001864e7f14

The concatenated balance appears as:
- (C) 0x0000000000000000000000000000000000187094dba0000000000001864e7f14

There is an 84-bit shift between each balance. 

Additionally, I introduced a constraint for the concatenated balances, formulated as \( C = (A \ll 83) + B \). However, it's not working as expected—there is an error when verifying the proof of commitment. It would be helpful if you could identify any missing components for the nonzero constraint in the plonkish backend in this PR.

I will convert this to a review once it is fixed.


sifnoc:
This is replaced by https://github.com/summa-dev/summa-solvency/pull/29

PR #294, opened 2024-06-04
Added concatenated balance for nonzero constraint process

sifnoc:
I added an additional advice column for the concatenated balance, which combines more than two balances into one with shifted positions. For example, consider the following total balances:
- (A) 0x00000000000000000000000000000000000000000000000000000003FFFFFFF
- (B) 0x00000000000000000000000000000000000000000000000000000002FFFFFFF
- (C) 0x00000000000000000000000000000000000000000000000000000001FFFFFFF

The concatenated balance appears as:
- (D) 0x00000000000003FFFFFFF00000000000002FFFFFFF00000000000001FFFFFFF
There is an 84-bit shift between each balance. 

Additionally, I introduced a constraint for the concatenated balances, formulated as \( D = (A \<\< (84 * 2) ) + (B \<\< 84) + C \)


prover/src/entry.rs
```diff
@@ -34,6 +34,18 @@ impl<const N_CURRENCIES: usize> Entry<N_CURRENCIES> {
...
     }
 
+    pub fn concatenated_balance(&self) -> BigUint {
+        let mut concatenated_balance = BigUint::from(0u32);
+
+        // To arragne the balances in the correct order, we need to reverse the array
+        for (i, balance) in self.balances.iter().rev().enumerate() {
+            // Shift bits: 1 for buffer; 19 is highest bit for two power of userbase; 64 is for the maximum range check limit
+            concatenated_balance += balance << ((1 + 19 + 64) * i);
+        }
```
alxkzmn, line 44: I think it's better to have `19` as a constant or a generic and also add an assertion that the concatenated balance would not exceed the field bit size.

prover/src/circuits/summa_circuit.rs
```diff
@@ -84,28 +80,68 @@ impl<
...
+            let mut current_shift = Expression::Constant(base_shift);
+
+            for i in (0..N_CURRENCIES - 1).rev() {
+                let balance = meta.query_advice(balances[i], Rotation::cur());
+                let shifted_balance = balance * current_shift.clone();
+                balances_expr = balances_expr + shifted_balance;
+
+                if i != 0 {
+                    current_shift = current_shift * Expression::Constant(base_shift);
+                }
```
alxkzmn, line 118: This works because we know exactly that `N_CURRENCIES=3`, but will break if the prover changes the `N_CURRENCIES`. I think it's better to add an assertion about the parameter combination (i.e., would there be "overlap" between shifted values for the given N_USERS, N_CURRENCIES and max possible range of a balance) and also calculate the `base_shift` based on these parameters.

alxkzmn (changes requested):

prover/src/circuits/summa_circuit.rs
sifnoc: While adding the assertions you suggested, I noticed that the verification of proof generated with `N_CURRENCIES` = 2 failed. I still don't have any clue why it doesn't work with two currencies. Therefore, I have simply modified it to allow only 1 or 3 currencies with the assertion.

I believe I can remove contract and fix the backend as I did in PR #297 after merging PR #292.

prover/src/utils/operation_helpers.rs
```diff
@@ -59,3 +59,87 @@ pub fn uni_to_multivar_binary_index<F: Field + PrimeField>(x: &usize, num_vars:
...
+    let user_base_bits = N_USERS.next_power_of_two().ilog2() as usize;
+
+    if user_base_bits > maximum_allowance_user_base_bits {
+        return Err(format!(
+            "The bit count for the user base exceeds the maximum limit of {}",
+            maximum_allowance_user_base_bits
+        ));
+    }
+
+    // Define shift bits: 1 for buffer, bits for user base that not exceed 19, and 64 bits for the balances range check
```
alxkzmn, line 83: Why 19?

prover/src/utils/operation_helpers.rs
```diff
@@ -59,3 +59,87 @@ pub fn uni_to_multivar_binary_index<F: Field + PrimeField>(x: &usize, num_vars:
...
 }
+
+pub fn calculate_shift_bits<const N_USERS: usize, const N_CURRENCIES: usize>(
+) -> Result<usize, String> {
+    // Define the maximum number of bits that can be used, based on the modulus in bn254.
+    const MAX_ALLOWANCE_BITS: usize = 253;
+
+    // Calculate the maximum number of bits that can be allocated to the user base,
+    // taking into account the number of currencies and the bits needed for the balances range check and buffer.
+    let maximum_allowance_user_base_bits = (MAX_ALLOWANCE_BITS / N_CURRENCIES) - 64 - 1;
```
alxkzmn, line 70: I'm confused about the 64 bits for the range check. If the range check decomposition is happening in separate advice columns, why do we need to reserve 64 bits here?

alxkzmn: I got it, 64 here means the max allowed bits for the user balance, which is related to range check indirectly :)

alxkzmn (approved):

prover/src/utils/operation_helpers.rs
sifnoc: Firstly, we can allocate 84 bits for each of the three currencies, totaling 252 bits, which does not exceed the modulus in the field.

The maximum user balance is $2^{64} - 1$, which is the maximum number that can be accommodated in the range check chip. After allocating 64 bits for the user balance, we are left with 20 bits.

We use one bit as a buffer to separate the total sum balance in the concatenated balance, leaving us with 19 bits at the end. That's why the user base shouldn't exceed $2^{19}$.

PR #296, opened 2024-07-11
Fixed V3 Backend

sifnoc:
Major Changes in V3 backend.
- Replaced the SRS file from `hermez-raw-#` to `hermez-raw-hyperplonk-#`, which is generated for the HyperPlonk Backend.
- Verifier parameters are now required in the process of verifying the proof instead of the 'verifying_key'.
- The `Round` interface in the Backend has been changed.
- The Summa Contract has been removed due to the absence of a Solidity verifier for proofs generated by the HyperPlonk backend.
- The GH Action Workflow no longer requires an AWS instance for testing.


sifnoc:
This is replaced by #297

PR #297, opened 2024-07-11
Fixed V3a Backend

sifnoc:
Major Changes in V3 backend.
- Replaced the SRS file from `hermez-raw-#` to `hyperponk-srs-#`, which is generated for the HyperPlonk Backend.
- Verifier parameters are now required in the process of verifying the proof instead of the 'verifying_key'.
- The `Round` interface in the Backend has been changed.
- The Summa Contract has been removed due to the absence of a Solidity verifier for proofs generated by the HyperPlonk backend.
- The GH Action Workflow no longer requires an AWS instance for testing.


backend/README.md
```diff
@@ -7,141 +7,112 @@ This directory contains the backend implementation for the Summa Proof of Solven
...
 Upon successful execution, you'll find a file named `user_0_proof.json` and see the following message:
 
+```bash
+2. Exported proof to user #0, as `user_0_proof.json`, with verifier params `verifier_params.json`
 ```
-3. Exported proof to user #0, as `user_0_proof.json`
-```
 
-### 4. Verify Proof of Inclusion
+Note that the `verifier_params.json` file can be used in any other round unless the same circuit configurations, such as `N_CURRENCIES` and `K`.
```
alxkzmn, line 101: Do you mean that it can be re-used if N_CURRENCIES and K didn't change?

prover/src/circuits/summa_circuit.rs
```diff
@@ -112,6 +112,13 @@ pub mod summa_hyperplonk {
                 grand_total,
             }
         }
+
+        pub fn init_empty() -> Self {
```
alxkzmn, line 116: Is this method used anywhere?

alxkzmn (approved):
LGTM!

backend/README.md
sifnoc: Yes, exactly!

prover/src/circuits/summa_circuit.rs
sifnoc: Thanks for pointing that out! I almost forgot to update its usage:

https://github.com/summa-dev/summa-solvency/blob/15374914a63878e92e7ee0f332ca2634496fef01/backend/examples/summa_solvency_flow.rs#L120-L126

This is how it can be used.

PR #301, opened 2024-09-06
(feature): Vk Verification

protocolwhisper:
Implement Validation Process for Verifying Key Contract

### DIFF
--- a/contracts/src/Summa.sol
+++ b/contracts/src/Summa.sol
@@ -5,10 +5,12 @@ pragma solidity ^0.8.18;
 //import "hardhat/console.sol";
 
 import "@openzeppelin/contracts/access/Ownable.sol";
+import "./Vkverifier.sol";
 
 import "./interfaces/IVerifier.sol";
 
-contract Summa is Ownable {
+
+contract Summa is Ownable , vkVerifier {
     /**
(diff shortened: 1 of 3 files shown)


sifnoc:
This can be used as an interface contract if someone (or you) implements the remaining verification process for the verifier key contract.

I would like to suggest that `vkVerifier` does not need to be inherited into the `Summa` contract.  The contract should be independent, as it can be used for any verifying key contract.

The `vkVerifier` should:
- Load data from the deployed verifier contract ([InclusionVerifier.sol](https://github.com/summa-dev/summa-solvency/blob/master/contracts/src/InclusionVerifier.sol) in V1 or [VerifyingKey.sol](https://github.com/summa-dev/summa-solvency/blob/v2/contracts/src/VerifyingKey.sol) in V2), with the contract address provided by the user. You can refer to this example: https://github.com/summa-dev/summa-solvency/blob/f11b1fe194cdde49d8d38030b45851c6829bb859/contracts/src/Summa.sol#L174
- The current `verifying_vk` method looks good, but it requires additional inputs to compare all elements of the keys, not just `vk_digest`, `k`, `n_inv`, and `omega`, but the rest of them as well.

I would be happy to accept your contribution if you modify the contract to be used as the final step in the validation process.

